PR #115, opened 2022-05-12
WalletConnect and wallet T&Cs

adavies1:
### You need some new `.env` variables for this stuff!

```
REACT_APP_WALLETCONNECT_INFURA_ID=<infura id>
REACT_APP_REQUIRED_ETHEREUM_NETWORK=kovan
REACT_APP_TOS_URL=https://app.voltz.xyz/tos
```

I have already added these to AWS, although the `infura_id` will need updating when we have a proper one. There is an issue where WalletConnect wallets don't return any wallet data atm, Costin is looking at it =)


arturbeg:
Hey Alex, what's the latest on the data retrieval issue with Wallet Connect? Do we need an infura id before we can merge this into production?

Sol3iman:
There is no data retrieval issue. We identified that the issue lies in the logic of how the position tables are displayed (or not) when you have no positions. 

When you have no open positions then the positions page just asks you to connect your wallet even if you are connected. This is an incorrect message. We can resolve that by changing the message to something like 'Please open a position'. For now this branch can be merged and the small fix will be pushed as a separate PR.

Sol3iman:
Need to sort out the infura ID for this. 

Do not merge.

Sol3iman:
I have created an Infura ID and added it to AWS so it can be used when this merge happens. 

Merge.

Sol3iman (approved):

PR #124, opened 2022-05-19
fix: corrected the fcm copy and logic

Sol3iman:
This is to fix the previous incorrectly merge PR which changed some FCM logic.


Sol3iman:
Do not merge

Sol3iman:
refactored so fcmMode is gone, up for review now.

PR #164, opened 2022-05-31
fix: change risky account copy

Sol3iman:


Sol3iman:
I think the copy needs to be changed to something a little more succinct and/or the design needs to change. This PR is WIP for now.

arturbeg:
yeah, not sure we can merge this to main since the text will just overflow the button @Sol3iman ?

Sol3iman:
> 

Yes, see comment above, please.

PR #165, opened 2022-05-31
fix: if db is down, user can log in without signing

Sol3iman:
Currently, if the db goes down people can't connect their wallets. 
This allows users to connect in the case of a db issue.


arturbeg:
hey @adavies1 are we ready to merge this?

PR #190, opened 2022-06-30
fix: update margin, leverage box, notional and margin units

Sol3iman:
Update margin now works
Leverage box next to slider now shows 5 digits + comma
Notional and margin in portfolio positions now show correct underlying token next to the value.
Put in placeholder for tooltips for stETH and leverage info circles.


avniculae (approved):
Looks good to me.

PR #234, opened 2022-08-31
fix: margin field in pending screen now shows 0 instead of NaN

Sol3iman:
- Pending screen margin field shows 0 instead of NaN now when margin is not defined.


src/components/containers/ConnectedMintBurnForm/ConnectedMintBurnForm.tsx
```diff
@@ -108,8 +109,9 @@ const ConnectedMintBurnForm: React.FunctionComponent<ConnectedMintBurnFormProps>
         transactionId={transactionId} 
         onComplete={handleComplete}
         notional={form.state.notional}
-        margin={Math.abs(form.state.margin as number) * (form.isRemovingMargin ? -1 : 1) }
-        onBack={handleGoBack} 
+        margin={isUndefined(form.state.margin && form.isRemovingMargin) ? 0 :
```
0xZenus, line 112: hmm, is `isUndefined(form.state.margin && form.isRemovingMargin)` intended (both in `isUndefined`)?

PR #251, opened 2022-09-19
Ir/google analytics

ioanaroceanu2000:
Added button IDs to track on GA
Buttons for pool table rows (trade/provide liquidity/edit/settle/rollover) are identified by:
- protocol name (e.g. cETH)
- isBorrow
- actor: trader/LP
Buttons inside a form (submit transaction/back/portfolio/view on etherscan) are identified by:
- protocol
- isBorrow
- margin action (Add/Remove)
- liquidity action
- notional Action
- agent: LP/VT/FT

* Back button in pending screen when a transaction has been declined from the wallet is also identified with a "FAILED" tag


src/components/interface/BorrowForm/components/SubmitControls/SubmitControls.tsx
```diff
@@ -13,7 +13,7 @@ interface SubmitControlsProps {
   isTradeVerified: boolean;
   onCancel: () => void;
   onSubmit: () => void;
-  protocol?: string;
+  gaButtonId?: string;
```
arturbeg, line 16: why did we get rid of protocol?

PR #258, opened 2022-09-26
Ir/rollover qa bugs

ioanaroceanu2000:
- updated SDK to branch `ir/eth-rollover-balance` which fixes ETH balance in rollover form and the ETH rollover margin bug
- added new pool icon to Rollover&Settlement pending transaction screens
- rollover trading information now shows min margin required in underlying (instead of yield-bearing token)
- fixed net notional "Loading...0" issue from portfolio when all positions are settled


arturbeg (approved):

arturbeg:
can we merge this into develop for now and then merge develop into main once happy with state of qa @ioanaroceanu2000 ?

PR #266, opened 2022-10-13
Save Terms of Service along with each signature AND LOCK USERS OUT IF THIS FAILS

CyclopsRex:
Please do not merge without coordination with other repos. See https://www.notion.so/voltz/T-Cs-2a70d816c4ba4256975ac900767da1eb#af01ec0e9f654824a21d29b5e2615f9a


ssbarbee (approved):
LGTM

PR #267, opened 2022-10-13
Save Terms of Service along with each signatures

CyclopsRex:
Please do not merge without coordination with other repos. See https://www.notion.so/voltz/T-Cs-2a70d816c4ba4256975ac900767da1eb#af01ec0e9f654824a21d29b5e2615f9a

### DIFF
--- a/src/contexts/WalletContext/services.ts
+++ b/src/contexts/WalletContext/services.ts
@@ -1,15 +1,14 @@
-import { WalletName, WalletRiskAssessment } from "./types";
+import { WalletName, WalletRiskAssessment } from './types';
 import detectEthereumProvider from '@metamask/detect-provider';
-import WalletConnectProvider from "@walletconnect/ethereum-provider";
-import { BigNumber, ethers, Contract } from "ethers";
-import { OverrideTypes } from "@utilities";
-import { DateTime } from "luxon";
+import WalletConnectProvider from '@walletconnect/ethereum-provider';
+import { ethers } from 'ethers';
 
 export type SignatureResponse = {
   signature?: string;
   timestamp?: string;
   walletAddress?: string;
-}
+  message?: string;
+};
 
 const unavailableText = 'Service unavailable, please try again shortly';
 
@@ -19,9 +18,13 @@ const unavailableText = 'Service unavailable, please try again shortly';
  */
 export const checkForCorrectNetwork = async (provider: ethers.providers.JsonRpcProvider) => {
   const network = await provider.getNetwork();
-  if(network.name !== process.env.REACT_APP_REQUIRED_ETHEREUM_NETWORK) {
+  if (network.name !== process.env.REACT_APP_REQUIRED_ETHEREUM_NETWORK) {
     // eslint-disable-next-line
-    console.warn(`User wallet is connected to '${network.name}' instead of '${process.env.REACT_APP_REQUIRED_ETHEREUM_NETWORK}'`);
+    console.warn(
+      `User wallet is connected to '${network.name}' instead of '${
+        process.env.REACT_APP_REQUIRED_ETHEREUM_NETWORK || '<unknown>'
+      }'`,
+    );
     throw new Error('Wrong network');
   }
 };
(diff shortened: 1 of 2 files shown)


src/contexts/WalletContext/services.ts
```diff
@@ -189,24 +232,26 @@ export async function getWalletRiskAssessment(walletId: string) {
     const result = await fetch('https://api.trmlabs.com/public/v2/screening/addresses', {
       method: 'POST',
       headers: {
-        'Accept': 'application/json',
+        Accept: 'application/json',
         'Content-type': 'application/json',
-        Authorization: 'Basic ' + btoa(process.env.REACT_APP_TRM_API_KEY || '')
+        Authorization: 'Basic ' + btoa(process.env.REACT_APP_TRM_API_KEY || ''),
```
ssbarbee, line 237:
```suggestion
        'Authorization': 'Basic ' + btoa(process.env.REACT_APP_TRM_API_KEY || ''),
```

keep consistency?

src/contexts/WalletContext/services.ts
```diff
@@ -32,76 +36,114 @@ export const checkForCorrectNetwork = async (provider: ethers.providers.JsonRpcP
...
  */
 export const getSignature = async (walletAddress: string) => {
   try {
-    const resp = await fetch(`https://voltz-rest-api.herokuapp.com/get-signature/${walletAddress}`, {
-      headers: {
-        'Content-Type': 'application/json',
-      }
-    });
+    const resp = await fetch(
+      `https://hackathon-rest-api.herokuapp.com/get-signature/${walletAddress}`,
```
ssbarbee, line 88: why do we hit the hackaton url?

CyclopsRex: Good spot - thanks! Fixed.

src/contexts/WalletContext/services.ts
CyclopsRex: If I make your suggested change, prettier (or linting; not sure which) undoes it automatically. (I guess the difference is the '-' in 'Content-type'.) So leaving this as is.

ssbarbee: lol :D gonna fix that one of these days! thanks for the headsup!

src/contexts/WalletContext/services.ts
```diff
@@ -32,66 +35,76 @@ export const checkForCorrectNetwork = async (provider: ethers.providers.JsonRpcP
...
+) => {
   const signerAddress = await signer.getAddress();
   const signatureData = await getSignature(signerAddress);
+  const latestTOS = getTOSText();
   let termsAccepted = false;
 
-  if(signatureData?.signature) {
-    const existingTOSSignerAddress = ethers.utils.verifyMessage(getTOSText(), signatureData.signature);
-    if(existingTOSSignerAddress === signerAddress) {
+  if (!!signatureData?.walletAddress && signatureData.walletAddress === signerAddress) {
```
ssbarbee, line 58:
```suggestion
  if (!!signatureData?.walletAddress && signatureData.walletAddress === signerAddress && signatureData.message === latestTOS) {
```

you can remove the nesting of 'ifs' by merging the 2 ifs.

src/contexts/WalletContext/services.ts
```diff
@@ -32,66 +35,76 @@ export const checkForCorrectNetwork = async (provider: ethers.providers.JsonRpcP
...
 
-  if(!termsAccepted) {
+  if (!termsAccepted) {
     try {
-      const signature = await signer.signMessage(getTOSText());
-      await saveSignature(signerAddress, signature);
-    } catch(e) {
-      throw new Error(' ');
+      // The latest terms of service have not been accepted. Get a signature now.
+      const termsOfService = getTOSText();
```
ssbarbee, line 69:
```suggestion
      const termsOfService = latestTOS; 
```

you already have it a bit above from what I can tell. Perhaps the variable termsOfService in this context is redundant even and we can use `latestTOS`?

src/contexts/WalletContext/services.ts
```diff
@@ -219,25 +235,31 @@ export async function getWalletRiskAssessment(walletId: string) {
  */
 export const isWalletRisky = (riskAssessment?: WalletRiskAssessment) => {
   const indicators = riskAssessment?.addressRiskIndicators;
-  const redFlag = Array.isArray(indicators) && indicators.find(i => i.categoryRiskScoreLevel >= 10);
+  const redFlag =
+    Array.isArray(indicators) && indicators.find((i) => i.categoryRiskScoreLevel >= 10);
```
ssbarbee, line 239:
```suggestion
    Array.isArray(indicators) && indicators.some((i) => i.categoryRiskScoreLevel >= 10);
```

perhaps more performant since it returns bool and not the item (object)

src/contexts/WalletContext/services.ts
CyclopsRex: Agreed that this is equivalent. I'm not actually sure that it's either more readable or any more performant (probably equivalent once compiled?) but less nesting is nice so I've made the suggested change.

src/contexts/WalletContext/services.ts
CyclopsRex: Open to this change but I don't think this PR is the right place cos I haven't touched this function (other than it being auto-formatted by prettier).

ssbarbee: Alrighty ^_^ I just saw it as opportunity for improvement. It won't break anything

ssbarbee:
hey Rex, I think all conversations in the PR look good. I am worried that there is a 1 failing check. Is this something you can also fix or it is expected to fail?

ssbarbee (approved):
LGTM

PR #268, opened 2022-10-13
Ir/ga v2

ioanaroceanu2000:
**SDK changes according to the last version of develop branch (same as main now)**
Changes:
- `session Id` recorded in the Wallet context so it can be used all over the app.
- send `dataLayer` events to GA when margin/notional/leverage/expectedApy change
- send 'page_view' event when page changes (including pending/failed/succ transaction screen)


src/utilities/googleAnalytics.ts
```diff
@@ -38,3 +47,53 @@ export const getNotionalActionFromHintState = (hint: SwapFormSubmitButtonHintSta
             return "";
     }
 }
+
+export const pushEvent = (eventName: string, eventValue: string | number, sessionId?: string, amm?: AugmentedAMM, agent?: string) => {
```
ssbarbee, line 51: eventName should a string literal type. For example we should limit what type of events we need to push.

```suggestion
type AnalyticEvents = 'a' | 'b' | 'c'...;

export const pushEvent = (eventName: AnalyticEvents, eventValue: string | number, sessionId?: string, amm?: AugmentedAMM, agent?: string) => {
```

src/components/interface/SwapForm/components/Leverage/Leverage.tsx
```diff
@@ -22,6 +24,10 @@ export type LeverageProps = {
 }
 
 const Leverage = ({availableNotional, minMargin, notional, onChange, value, resetDeltaState}: LeverageProps) => {
+  const { sessionId } = useWallet();
```
ssbarbee, line 27: remove these vars. probably some leftovers

src/utilities/googleAnalytics.ts
ssbarbee, line 51:
```suggestion
type AnalyticEventPayload = {
 eventName: string;
 eventValue: string | number; 
 sessionId?: string;  // check if this can be required param
 amm?: AugmentedAMM;
 agent?: string; // check if we can limit the agent to certain values, similar to the eventName
}
export const pushEvent = ({ eventName, eventValue, sessionId, amm, agent }: AnalyticEventPayload) => {
```

src/contexts/MintBurnFormContext/MintBurnFormContext.tsx
```diff
@@ -267,6 +268,26 @@ export const MintBurnFormProvider: React.FunctionComponent<MintBurnFormProviderP
...
+  const notionalChange = useCallback(debounce((value: number | undefined) => {
+    pushEvent("notional_change", value ?? 0, sessionId, poolAmm, "Liquidity Provider");
+  }, 1000), []);
+
+  const marginChange = useCallback(debounce((value: number | undefined) => {
+    pushEvent("margin_change", value ?? 0, sessionId, poolAmm, "Liquidity Provider");
+  }, 1000),[]);
+
+  useEffect(() => {
+    if ( !isUndefined(mintMinimumMarginRequirement.result)) {
```
ssbarbee, line 280: can you add a task in the FE UI where we can refactor the isUndefined to support null. I feel that we might be inviting bugs if someone passes null due to the component props allowing it.

src/contexts/WalletContext/WalletProvider.tsx
```diff
@@ -17,6 +17,12 @@ const WalletProvider: React.FunctionComponent<WalletProviderProps> = ({
   const [name, setName] = useState<WalletName | null>(null);
   const [balance, setBalance] = useState<Record<string, BigNumber>>({});
   const [required, setRequired] = useState<boolean>(false);
+  const [sessionId, setSessionId] = useState<string>("");
+
+  useEffect(() => {
+    setSessionId((Math.random()).toString(20).substring(2, 10));
```
ssbarbee, line 23: lets use the package uuid from npm. yarn add -E uuid@9.0.0. https://www.npmjs.com/package/uuid 

then simple use their utility to generate the random id.

src/utilities/googleAnalytics.ts
```diff
@@ -1,8 +1,17 @@
...
 import { isUndefined } from "lodash";
 import AugmentedAMM from "./augmentedAmm";
 import isBorrowing from "./isBorrowing";
 
+declare global {
+    interface Window {
+      dataLayer?: any;
+    }
+}
+
```
ssbarbee, line 14: lets add the sessionId in this file.

for example

const SESSION_ID = uuid();

and then we can use it in our utility for tracking. The idea being is that whenever the user refreshes the site we generate 1 session id for tracking

src/utilities/googleAnalytics.ts
```diff
@@ -38,3 +47,53 @@ export const getNotionalActionFromHintState = (hint: SwapFormSubmitButtonHintSta
             return "";
     }
 }
+
+export const pushEvent = (eventName: string, eventValue: string | number, sessionId?: string, amm?: AugmentedAMM, agent?: string) => {
+    // with context
+    if (amm) {
+        const pool = amm.protocol + (amm.rateOracle.protocolId === 5 || amm.rateOracle.protocolId === 6 ? "_borrow" : "" );
+        window.dataLayer.push({'event': eventName, 'value': eventValue, 'sessionId': sessionId, 'pool': pool, 'agent': agent});
```
ssbarbee, line 55: maybe check if window.dataLayer is present at the top of the function before actually using it.

something like

if(!window.dataLayer) {
  return;
}

src/utilities/googleAnalytics.ts
```diff
@@ -38,3 +47,53 @@ export const getNotionalActionFromHintState = (hint: SwapFormSubmitButtonHintSta
...
+    } else { // without context
+        window.dataLayer.push({'event': eventName, 'value': eventValue, 'sessionId': sessionId});
+    }
+}
+
+export const pushTxSubmition = (
```
ssbarbee, line 61: maybe check if window.dataLayer is present at the top of the function before actually using it.

something like

if(!window.dataLayer) {
  return;
}

src/utilities/googleAnalytics.ts
```diff
@@ -38,3 +47,53 @@ export const getNotionalActionFromHintState = (hint: SwapFormSubmitButtonHintSta
...
+    eventName: string, 
+    notional: number | undefined,
+    margin: number | undefined,
+    action: string,
+    sessionId: string,
+    amm: AugmentedAMM,
+    agent?: string | number,
+    failMessage?: string
+) => {
+    let agentType = "";
```
ssbarbee, line 71: if possible all mapping logic should be extracted away from the core utilities. The mapping is a logic closer to the business layer, not the tracking layer. We should avoid mixing these logics to prevent tech debt.

src/utilities/page.ts
```diff
@@ -1,7 +1,10 @@
+import { pushEvent } from "./googleAnalytics";
+
 /**
  * Allows you to set the suffix of the page title
  * @param text - the suffix to set on the page title
  */
 export const setPageTitle = (text: string) => {
-  document.title = `Voltz Protocol ⚡ The Interest Rate Swap AMM${text ? ` - ${text}` : ''}`;
+  document.title = `Voltz ⚡ ${text ? `${text}` : ''}`;
+  pushEvent("path_change", text);
```
ssbarbee, line 9: proposal: lets change this event name to be title_change.

Let's add a track page change event (path_change) inside the App.tsx. We can use the API offered to us by react-router. Ping me for more details :D internet is slow...

ssbarbee, line 9: proposal: lets change this event name to be title_change.

Let's add a track page change event (path_change) inside the App.tsx. We can use the API offered to us by react-router. Ping me for more details :D internet is slow...

src/contexts/WalletContext/WalletProvider.tsx
ssbarbee: @ioanaroceanu2000 
Why was this resolved? 
Can you share the commit where this was addressed? 
If you have it locally push it please!

Best practice is to resolve the comment after you pushed the fix. The share it via message in the conversation. 

Example message: "@reviewer fixed in <commit_hash>"

src/components/interface/SwapForm/components/Leverage/Leverage.tsx
ssbarbee: @ioanaroceanu2000 
Why was this resolved? 
Can you share the commit where this was addressed? 
If you have it locally push it please!

Best practice is to resolve the comment after you pushed the fix. The share it via message in the conversation. 

Example message: "@reviewer fixed in <commit_hash>"

src/utilities/page.ts
ioanaroceanu2000: changed event name at [commit](https://github.com/Voltz-Protocol/voltz-ui/pull/268/commits/9bb0958572bdde864d23200e7f12fd6c19910057). Let's talk about actual path change, this would be very useful

src/utilities/googleAnalytics.ts
ioanaroceanu2000: created a new util function `getAgent.ts` and used it in the component logic at [commit](https://github.com/Voltz-Protocol/voltz-ui/pull/268/commits/9bb0958572bdde864d23200e7f12fd6c19910057)

src/utilities/googleAnalytics.ts
ioanaroceanu2000: [checked](https://github.com/Voltz-Protocol/voltz-ui/pull/268/commits/9bb0958572bdde864d23200e7f12fd6c19910057)

src/utilities/googleAnalytics.ts
ioanaroceanu2000: duplicate of below

src/utilities/googleAnalytics.ts
ioanaroceanu2000: added
```
import { v4 as uuidv4 } from 'uuid';

const SESSION_ID = uuidv4()
```

src/contexts/WalletContext/WalletProvider.tsx
ioanaroceanu2000: resolved 9bb0958572bdde864d23200e7f12fd6c19910057

src/contexts/MintBurnFormContext/MintBurnFormContext.tsx
ioanaroceanu2000: I'm adding this to a new ticket bc it needs a lot of refactoring that's out of scope for this PR

src/utilities/googleAnalytics.ts
ioanaroceanu2000: created two new types `DataLayerEventPayload` and `TxEventPayload` to be used as parameters for the `pushEvent` function. This funct is now general, used for tx events, input change events, title change etc... resolved 9bb0958572bdde864d23200e7f12fd6c19910057

src/components/interface/PendingTransaction/PendingTransaction.tsx
```diff
@@ -2,17 +2,17 @@ import React, { useEffect, useMemo, useRef, useState } from 'react';
...
 import { AMMProvider, MintBurnFormLiquidityAction, useAMMsContext } from '@contexts';
-import { Button, Panel, Typography, Loading, TokenAndText } from '@components/atomic';
+import { Button, Panel, Typography, Loading } from '@components/atomic';
 import { ProtocolInformation, WalletAddressDisplay } from '@components/composite';
 import { formatCurrency } from '@utilities';
 import { isUndefined } from 'lodash';
 import { Position } from '@voltz-protocol/v1-sdk/dist/types/entities';
 import { Wallet } from '@graphql';
-import { ChangeCircleSharp } from '@mui/icons-material';
+import { getAgentFromPosition } from 'src/utilities/getAgent';
```
ssbarbee, line 15: move this to @utilities import, you might have failing tests

ssbarbee:
please resolve the conflicts and apply the small comment I have. Otherwise LGTM

ssbarbee (approved):

PR #271, opened 2022-10-18
Leaderboard without referrals

ssbarbee:
This PR brings the leaderboard. It seems @CyclopsRex removed the referral logic from the PR.

@arturbeg can you have a look here if that is the case?


src/components/interface/RankingTable/components/RankingTableHeader/RankingUserSummary.tsx
```diff
@@ -0,0 +1,362 @@
...
+            </TableBody>
+          </Table>
+        </TableContainer>
+      </Box>
+    );
+  };
+
+  const generateLink = (w: Wallet) => {
+    const address = w.account;
+    return !isNull(address) ? 'http://localhost:3000/#/trader-pools?invitedBy=' + address : null;
```
ssbarbee, line 219: @arturbeg is this the referal logic?

PR #272, opened 2022-10-19
Feature/add missing badge assets + connect SDK

ssbarbee:
This PR adds the missing badges for the Profile screen. Also creates a BE connection for Data


src/hooks/useCurrentSeason.ts
```diff
@@ -0,0 +1,27 @@
+import { DateTime } from 'luxon';
+
+const seasons = new Map<number, DateTime>();
+const SEASON_START_DATE = DateTime.fromISO('2022-08-01T16:03:44+00:00');
+const SEASON_DURATION = 1; // months
```
arturbeg, line 5: the duration should be one quarter (on average 3 months)

arturbeg: i'd probably just have a mapping with season start timestamp and season end timestamp that we can configure and set ourselves, wdyt?

src/hooks/useRanking.ts
```diff
@@ -1,98 +1,77 @@
...
-      seasonEndDate,
-      findCurrentSeason
-    ]);
-}
+  return useMemo(
+    () =>
+      ({
+        rankings,
+        claimStatic,
+        claim,
```
arturbeg, line 70: does the leaderboard have support for claiming? it is not supposed to at this stage since SBTs will roll out in phase 2

src/routes/Profile/CollectionBadge/CollectionBadge.test.tsx
```diff
@@ -16,7 +16,7 @@ test('renders proper UI when achieved', () => {
 
   expect(screen.getByTestId('CollectionBadge-leverageCrowbar')).not.toBeNull();
   expect(screen.getByText('LEVERAGE CROWBAR')).not.toBeNull();
-  expect(screen.getByText('Achieved: 10/12/22')).not.toBeNull();
+  expect(screen.getByText('Achieved: 02/02/22')).not.toBeNull();
```
arturbeg, line 19: why is this date hardcoded?

ssbarbee: the tests are done against a storybook that uses mock data. So we are asserting if the rendering will happen correctly. I am using 02/02 on purpose since I try to format a timestamp into locale of the machine :D (the server where tests are run seems to have different locale than my local one so 02/02, 03/03... etc work)

src/hooks/useCurrentSeason.ts
ssbarbee: fixed in: 075cc4da9fa1be34e1cb2de394276557873761ca. Check useCurrentSeason/constants.ts

src/hooks/useRanking.ts
ssbarbee: fixed in: 075cc4da9fa1be34e1cb2de394276557873761ca

PR #276, opened 2022-10-20
feat: user sees max withdrawable margin

Sol3iman:
- User should see how much they can actually withdraw from their margin position when trying to unwind a position or cash out.
- WIP as the data retrieval for the margin amount should ideally happen from the PortfolioContext but it is returning undefined for the margin information @ssbarbee let us review this together.

### DIFF
--- a/src/components/composite/MarginAmount/MarginAmount.tsx
+++ b/src/components/composite/MarginAmount/MarginAmount.tsx
@@ -6,6 +6,7 @@ import MaskedIntegerField from '../MaskedIntegerField/MaskedIntegerField';
 import InputTokenLabel from '../InputTokenLabel/InputTokenLabel';
 import { formatCurrency, toUSFormat } from '@utilities';
 import { HealthFactorText } from '@components/composite';
+import { useSwapFormContext, usePortfolioContext, usePositionContext } from '@contexts';
 
 export type MarginAmountProps = {
   balance?: number;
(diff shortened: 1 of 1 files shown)


src/components/composite/MarginAmount/MarginAmount.tsx
```diff
@@ -38,7 +39,23 @@ const MarginAmount: React.FunctionComponent<MarginAmountProps> = ({
...
+  const positionForm = usePositionContext();
 
+  let marginAmount;
+  if (underlyingTokenName?.includes('USD')) {
+    const portfolioMarginInfo = positionForm.position?.margin.toString();
+    marginAmount = !isUndefined(portfolioMarginInfo) ? parseInt(portfolioMarginInfo)/1e6 : defaultMargin;
+  } else {
+    const portfolioMarginInfo = positionForm.position?.margin.toString();
+    marginAmount = !isUndefined(portfolioMarginInfo) ? parseInt(portfolioMarginInfo) / 1e18 : defaultMargin;
+  }
```
ssbarbee, line 53:
```suggestion
  const divisor = underlyingTokenName?.includes('USD') ? 1e6 : 1e18;
  const portfolioMarginInfo = positionForm.position?.margin.toString();
  const marginAmount = !isUndefined(portfolioMarginInfo) ? parseInt(portfolioMarginInfo)/divisor : defaultMargin;
```

Reduce duplication of code 😁

ssbarbee (changes requested):

src/components/composite/MarginAmount/MarginAmount.tsx
Sol3iman: I think that works well and the withdrawable margin amounts seems reasonable. Happy to merge if you give green light. @ssbarbee

Sol3iman: Yes, we checked this just now in our peer review and the information seems to match up. Worth making a ticket for unifying the sources of margin information as currently PortfolioContext and PositionContext serve very similar information.

ssbarbee (approved):
LGTM

PR #280, opened 2022-10-24
fix: when disconnect button is pressed, the right flow is invoked

Sol3iman:
- Wallet disconnect button now correctly disconnects the wallet


src/components/interface/WalletConnect/components/WalletConnectModal/WalletConnectModal.tsx
```diff
@@ -19,7 +20,11 @@ const WalletConnectModal: React.FunctionComponent = () => {
   };
   const handleSelectWallet = (name: WalletName) => {
     handleClose();
-    wallet.connect(name);
+    if (walletOptions[2].name === name) {
```
ioanaroceanu2000, line 23: can we do "disconnect" === name? (or whatever the name is for that option). Otherwise, when we add other options we might forget to change the '2' index and get confused

Sol3iman: True, let's make it more human readable

ioanaroceanu2000 (approved):

PR #283, opened 2022-10-24
Ir/wallet network fix

ioanaroceanu2000:
* Catch this error and only show `wrong network` instead of the raw message 
Screenshot 2022-10-24 at 17 45 56
* Also do live UI updates on network/account change with Metamask

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,6 +5,8 @@ import { routes, TradingLeague, Profile, LiquidityProvider, Trader, FixedBorrowe
 import { AlphaBanner, GweiBar } from '@components/composite';
 import Box from '@mui/material/Box';
 import { useEffect } from 'react';
+import useWallet from './hooks/useWallet';
+import { ethers } from 'ethers';
 
 const App = () => {
   useEffect(() => {
@@ -12,6 +14,22 @@ const App = () => {
       TagManager.initialize({ gtmId: process.env.REACT_APP_GTM_CODE });
     }
   }, []);
+
+  const wallet = useWallet();
+
+  /* eslint-disable @typescript-eslint/no-unsafe-call */
+  /* eslint-disable @typescript-eslint/no-unsafe-member-access */
+  window.ethereum?.on('accountsChanged', function () {
+    if (!wallet.loading) {
+      void wallet.connect('metamask');
+    }
+  })
+
+  window.ethereum?.on('chainChanged', function () {
+    if (!wallet.loading) {
+      void wallet.connect('metamask');
+    }
+  })
 
   return (
     <>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,6 +16,7 @@ const App = () => {
 
   const wallet = useWallet();
 
+  // TODO upgrade to metamask v2.0.0 and remove ethereum.d.ts
   /* eslint-disable @typescript-eslint/no-unsafe-call */
   /* eslint-disable @typescript-eslint/no-unsafe-member-access */
   window.ethereum?.on('accountsChanged', function () {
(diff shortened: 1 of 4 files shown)


Sol3iman:
Can you explain the steps involved to reproduce this error? Should I simply connect with the right network and then while connected, switch to the wrong network to test the happy path? @ioanaroceanu2000

src/App.tsx
```diff
@@ -5,6 +5,8 @@ import { routes, TradingLeague, Profile, LiquidityProvider, Trader, FixedBorrowe
 import { AlphaBanner, GweiBar } from '@components/composite';
 import Box from '@mui/material/Box';
 import { useEffect } from 'react';
+import useWallet from './hooks/useWallet';
```
ssbarbee, line 8:
```suggestion
import { useWallet } from '@hooks';
```

lets keep same style throughout the code

src/contexts/WalletContext/services.ts
```diff
@@ -18,14 +18,18 @@ const unavailableText = 'Service unavailable, please try again shortly';
...
+  try {
+    const network = await provider.getNetwork();
+    if (network.name !== process.env.REACT_APP_REQUIRED_ETHEREUM_NETWORK) {
+      // eslint-disable-next-line
+      console.warn(
+        `User wallet is connected to '${network.name}' instead of '${
+          process.env.REACT_APP_REQUIRED_ETHEREUM_NETWORK || '<unknown>'
+        }'`,
+      );
+      throw new Error();
```
ssbarbee, line 30:
```suggestion
      throw new Error(`User wallet is connected to '${network.name}' instead of '${
          process.env.REACT_APP_REQUIRED_ETHEREUM_NETWORK || '<unknown>'
        }'`);
```

remove not needed console.warn and eslint-disable-next-line... Enrich the Error you throw

src/types/ethereum.d.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/// <reference types="react-scripts" />
+
+interface Window {
```
ssbarbee, line 4: Why is this needed? Please update the already existing global.d.ts file, with the new methods to added in App.tsx

src/App.tsx
```diff
@@ -13,6 +15,22 @@ const App = () => {
     }
   }, []);
 
+  const wallet = useWallet();
+
+  /* eslint-disable @typescript-eslint/no-unsafe-call */
```
ssbarbee, line 20: why eslint-disable?

please update type definitions in global.d.ts

src/App.tsx
```diff
@@ -13,6 +15,22 @@ const App = () => {
...
+  /* eslint-disable @typescript-eslint/no-unsafe-member-access */
+  window.ethereum?.on('accountsChanged', function () {
+    if (!wallet.loading) {
+      void wallet.connect('metamask');
```
ssbarbee, line 24: can we actually connect with different wallet other than metamask? Why do we hardcode metamask here? Please explain 🙏

ioanaroceanu2000:
> Can you explain the steps involved to reproduce this error? Should I simply connect with the right network and then while connected, switch to the wrong network to test the happy path? @ioanaroceanu2000

@Sol3iman it's very difficult to test, it occurs randomly. To force it, you need to specify the network in the provider instantiation and then try connecting with another network.

src/App.tsx
ioanaroceanu2000: `window.ethereum?.on` listens to metamask events only. So it means a change happened while connected to metamask. We use WalletConnect as well but the WalletConnect flow is not affected.

src/App.tsx
ioanaroceanu2000: done in ff692256e500f567310f338bb4a684c73a01433a

src/contexts/WalletContext/services.ts
ioanaroceanu2000: done in ff692256e500f567310f338bb4a684c73a01433a

src/types/ethereum.d.ts
ioanaroceanu2000: I added a TODO for metamask 2.0.0. This will solve the issue but let's not risk it before release

ssbarbee (approved):
LGTM

PR #293, opened 2022-10-25
Sol/leverage buttons

Sol3iman:
- Remove leverage slider and replace with leverage buttons
- Buttons are preset with 100x, 500x, 1000x
- The 100x button is preselected

### DIFF
--- a/src/components/interface/SwapForm/components/Leverage/Leverage.tsx
+++ b/src/components/interface/SwapForm/components/Leverage/Leverage.tsx
@@ -1,7 +1,6 @@
 import React, { useEffect, useRef, useState } from 'react';
 import Box from '@mui/material/Box';
 import { MaskedIntegerField, IconLabel } from '@components/composite';
-import { colors } from '@theme';
 import { isNumber, isUndefined } from 'lodash';
 import { formatNumber, notFormatted, stringToBigFloat, toUSFormat } from '@utilities';
 import { Button } from '@mui/material';
@@ -101,7 +100,7 @@ const Leverage = ({
         {options.map(opt => {
           return (
             <Button
-              variant={"contained"} //
+              variant={"contained"}
               size ="small"
               color="secondary"
               sx={opt.id === activeOption.id ? activeButtonStyle : buttonStyle}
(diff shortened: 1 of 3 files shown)


src/components/interface/SwapForm/components/Leverage/Leverage.tsx
```diff
@@ -124,56 +90,29 @@ const Leverage = ({
...
-              height: '6px',
-              width: '3px',
-              background: isDisabled ? undefined : colors.vzCustomRed2.base,
-            },
-          }}
-        />
+        {options.map(opt => {
+          return (
+            <Button
+              variant={"contained"}
```
ssbarbee, line 103:
```suggestion
              key={opt.id}
              variant={"contained"}
```

add the key prop, required by React when enumerating a list

src/components/interface/SwapForm/components/Leverage/Leverage.tsx
```diff
@@ -44,49 +44,15 @@ const Leverage = ({
...
-  const rainbow3 = !isDisabled ? Math.max(high / 1.2, 1) : 17;
-  const rainbow4 = !isDisabled ? Math.max(high / 1.05, 1) : 19;
-
-  const rainbow2Percent = (rainbow2 / high) * 100;
-  const rainbow3Percent = (rainbow3 / high) * 100;
-  const rainbow4Percent = (rainbow4 / high) * 10;
-
-  const rainbowStart = Math.max((1 - rainbow1 / range) * 100, 0);
-  const rainbowWidth = 100 - rainbowStart;
+  const options = [{ id: 100, text: "100x" }, { id: 500, text: "500x" }, { id: 1000, text: "1000x" }];
```
ssbarbee, line 47: perhaps better structure for the array is just array of numbers since the text is computed by appending 'x' to the number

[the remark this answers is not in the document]
Sol3iman: Moved it to outside the comp definition.

src/components/interface/SwapForm/components/Leverage/Leverage.tsx
Sol3iman: Included the key prop as key={opt} since I removed the id variable and left it as an array of numbers

src/components/interface/SwapForm/components/Leverage/Leverage.tsx
Sol3iman: Turned it into just an array of numbers without text and id. 
Simpler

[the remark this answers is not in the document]
ssbarbee: reason is not to allocate new arrays on every render. You can also do that using react.memo but I kinda don't like to have yet another fn call when you can move it out.

src/components/interface/SwapForm/components/Leverage/Leverage.tsx
```diff
@@ -44,49 +46,14 @@ const Leverage = ({
     isUndefined(internalValue);
   const timer = useRef<number>();
 
-  const maxNotional = !isDisabled ? Math.min(notional, availableNotional) : 10;
```
ssbarbee, line 47: seems like we removed the usages of the variable isDisabled. IMO it is now unused variable. Can you delete it?

src/components/interface/SwapForm/components/Leverage/Leverage.tsx
```diff
@@ -123,57 +90,31 @@ const Leverage = ({
       <Box
         sx={{
```
ssbarbee, line 91: you can move this style object to the style.ts file. Keep the logic component separate from styling

Sol3iman: Moved it to the style file

ssbarbee (approved):

PR #296, opened 2022-10-27
feature/introduce styled

ssbarbee:
Template the way we do FE work.
Styling should be done via styled utilities from material UI which enable us to have a dedicated files for the style, it also allow us to have scalable components that we can append styles to.


ioanaroceanu2000 (approved):
nice

PR #298, opened 2022-10-28
feat: send user's address as GA event param

ioanaroceanu2000:
When user connects, save their address and send it as an event param to GA


src/utilities/googleAnalytics.ts
```diff
@@ -10,6 +10,11 @@ import isBorrowing from './isBorrowing';
 import { v4 as uuidv4 } from 'uuid';
 
 const SESSION_ID = uuidv4();
+let USER_ADDRESS = '';
+
+export const saveUserAddress = (address: string) => {
+  USER_ADDRESS = address;
```
arturbeg, line 16: is there a cleaner way to cache the user address vs. using a global variable? wdyt @ssbarbee @ioanaroceanu2000 ?

src/utilities/googleAnalytics.ts
```diff
@@ -90,5 +95,5 @@ export const pushEvent = (payload: DataLayerEventPayload) => {
   if (!window.dataLayer) {
     return;
   }
-  window.dataLayer.push({ sessionId: SESSION_ID, ...payload });
+  window.dataLayer.push({ sessionId: SESSION_ID, userAddress: USER_ADDRESS, ...payload });
```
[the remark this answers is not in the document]
ioanaroceanu2000: "A data layer is a JavaScript object that is used to pass information from your website to your Tag Manager container. You can then use that information to populate variables and activate triggers in your tag configurations." 

Yes, for each event, the user's address is passed along other event-specific params

src/utilities/googleAnalytics.ts
ioanaroceanu2000: the other way is to use the `WalletContext` which can be called only in React components but this requires code change in all components that send a GA event. Filip's opinion might help

ssbarbee: best way is to include the WALLET address as required param in the pushEvent function. This way we remove the "state" from GA since it if you think about it the requirement states: "publish the user address to GA" (or translated extend the GA utility with param address)

ssbarbee: if you have places where you cannot use the "useWallet()" you can read from localStorage we should be keeping the address there

src/utilities/googleAnalytics.ts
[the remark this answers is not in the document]
arturbeg: got it, thank you!

src/utilities/googleAnalytics.ts
arturbeg: yeah, i also think we should aim to remove state from ga

ssbarbee: State is evil... it makes everything  more complex to test and refactor later on

ioanaroceanu2000: 8eb29680a912799f4984988c57ad166a8d269a11 commit with proposed param. I basically used the wallet hook wherever the `pushEvent()` function was called.

ssbarbee (approved):
Lgtm

PR #299, opened 2022-10-29
fix: prevent risky wallet error msg being overridden

Sol3iman:
- Prevent risky wallet error message being overridden by the general error msg in ProviderWrapper


ssbarbee (approved):
Is it not possible to do the verification of message based on some status? Approving now but please consider that hardcoded strings might get changed

Sol3iman:
> Is it not possible to do the verification of message based on some status? Approving now but please consider that hardcoded strings might get changed

This would probably be a better way but for now I’d like to just see if the errors are shown properly. We should make a ticket for this. Thanks for approval, will merge to develop now.

PR #307, opened 2022-10-31
fix: reset ecosystem on every location change when not in PAGE mode

ssbarbee:
had to dig a bit deep for this one :D


src/routes/Ecosystem/Ecosystem.tsx
```diff
@@ -56,6 +58,13 @@ const Ecosystem: React.FunctionComponent = () => {
     }
   }, [setPageTitle, renderMode]);
 
+  useEffect(() => {
```
arturbeg, line 61: so the logic is that the location.key is expected to change after I click on the "products" route of "ecosystem"? if yes then how's the key determined?

arturbeg (approved):

src/routes/Ecosystem/Ecosystem.tsx
ssbarbee: from what I read in the code of react-router is that every time there is a Link click (a component from react-router, that we use in the NavLink)  the location.key changes. react-router will prevent the page from refreshing, but if you listen to changes of the location.key it will be different and you can "opt in" to refresh the page this way (or do any other logic really).
location.key is something they autogenerate for every "route" render or request for a route to be rendered

PR #314, opened 2022-11-01
fix: remove not needed renders for the static component Nav

ssbarbee:
Using React.memo makes sense for this component since it is just static one. Look the before and after videos

https://user-images.githubusercontent.com/11079632/199205970-c58d7a04-8d1c-4dc7-a20a-e07069807939.mov

https://user-images.githubusercontent.com/11079632/199205950-1a045da8-771d-4921-bd23-d018bdb6103a.mov

### DIFF
--- a/src/components/interface/Nav/Nav.tsx
+++ b/src/components/interface/Nav/Nav.tsx
@@ -6,7 +6,7 @@ import { NavLink } from './NavLink/NavLink';
 import { iconSx } from './NavLink/style';
 import { colors } from '@theme';
 
-const Nav: React.FunctionComponent = () => (
+const Nav: React.FunctionComponent = React.memo(() => (
   <Box
     sx={{
       display: 'flex',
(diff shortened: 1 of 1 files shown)


src/components/interface/Nav/Nav.tsx
```diff
@@ -74,32 +74,31 @@ const Nav: React.FunctionComponent = () => (
...
-              text: 'FIXED BORROW',
-              link: `/${routes.BORROW_POS}`,
-            },
-          ]
-        }
-        
-      >
-        Ecosystem
-      </NavLink>
+    <NavLink
```
ioanaroceanu2000, line 77: why does this change keep coming up?

ioanaroceanu2000: maybe because w e need a rebase? "This branch is out-of-date with the base branch"

ioanaroceanu2000 (approved):

PR #318, opened 2022-11-03
refactor: performance of banner

ssbarbee:
Current solution was using direct DOM element manipulation using JS (abusing React). This is slow and CPU heavy. We should rely on CSS animation to get the stuff done.


src/components/composite/AlphaBanner/AlphaBanner.styled.tsx
```diff
@@ -0,0 +1,54 @@
...
+  from { 
+    transform: translateX(0); 
+  }
+  to { 
+    transform: translateX(-50%); 
+  }
+`;
+
+export const BannerContainer = styled(Box)`
+  background: linear-gradient(90deg, #201b35 0%, #372d65 50.52%, #2f2a54 100%);
```
ioanaroceanu2000, line 17: I think we had issues with this in the past. It would make the UI load really slow on chrome with 'hardware acceleration' turned off. Let's check if it has the same effect in that context

ssbarbee: Yes check it please. I think problem was impl. of prev solution 😆

.github/CODEOWNERS
```diff
@@ -4,4 +4,4 @@
 # These owners will be the default owners for everything in
 # the repo. Below owners, will be requested for
 # review when someone opens a pull request.
-*       @arturbegyan @ssbarbee @ioanaroceanu2000
+*       @arturbeg @ssbarbee @ioanaroceanu2000
```
arturbeg, line 7: oh good catch, funny i didn't notice it last time:)

arturbeg (approved):

PR #319, opened 2022-11-03
fix: add caching for fetching badges

ssbarbee:
Add caching for the profile when jumping between different seasons


src/graphql/getters/getSeasonBadges/helpers.ts
```diff
@@ -1,82 +1,26 @@
...
+export function getDefaultResponse(seasonId: Season['id']): GetProfileBadgesResponse {
+  return SEASON_BADGE_VARIANTS[seasonId].map((b) => ({
+    variant: b as BadgeVariant,
+  }));
+}
+
+/**
+ * "Convert seconds to milliseconds, but only if the input is a number and not zero."
+ *
+ * @param {number} seconds - number - The number of seconds to convert to milliseconds.
```
arturbeg, line 14: nice comments! very helpful

src/graphql/getters/getSeasonBadges/mappers.ts
```diff
@@ -0,0 +1,82 @@
+import { BadgeVariant } from './types';
+import { Season } from '../../../hooks/season/types';
+
+export const BADGE_TYPE_BADGE_VARIANT_MAP: Record<string, BadgeVariant> = {
+  // season 1
+  '12': 'fixedTrader',
```
arturbeg, line 6: are these id's for each badge from the BE?

src/routes/Profile/Profile.tsx
```diff
@@ -35,11 +37,24 @@ const Profile: React.FunctionComponent = () => {
...
   const getBadges = async (seasonId: Season['id'], account: string) => {
-    setLoading(true);
-    const result = await getSeasonBadges(account, seasonId);
+    const seasonUserId = getSeasonUserId(account, seasonId);
+    let result = getCacheValue(seasonUserId);
+
+    if (!result) {
+      setLoading(true);
+      result = await fetchBadges(seasonId, account);
+      setCacheValue(seasonUserId, result);
```
arturbeg, line 54: so this function effectively caches the badges so that we don't need to retrieve the badges next time the user wants to renavigate to the profile page as long as they don't reload the app?

src/routes/Profile/Profile.tsx
```diff
@@ -69,13 +85,10 @@ const Profile: React.FunctionComponent = () => {
...
-        const copy = [...prev];
-        copy.forEach((b) => {
-          if (b.variant === variant) {
-            b.claimedAt = Date.now().valueOf();
-          }
-        });
-        return copy;
+        return prev.map((b) => ({
+          ...b,
+          claimedAt: b.variant === variant ? Date.now().valueOf() : b.claimedAt,
```
arturbeg, line 90: trying to wrap my head around why claimedAt attribute is conditional on the variant

src/routes/Profile/Profile.tsx
```diff
@@ -102,6 +117,7 @@ const Profile: React.FunctionComponent = () => {
 
   useEffect(() => {
     setPageTitle('Profile', wallet.account);
+    invalidateCache();
```
arturbeg, line 120: why do we invalidate cache in here?

src/routes/Profile/Profile.tsx
```diff
@@ -84,12 +97,14 @@ const Profile: React.FunctionComponent = () => {
...
   }
-  // mocking utils
+  // end of mocking utils
+
   useEffect(() => {
     if (!wallet.account) {
       return;
     }
     void fetchEnsDetails(wallet.account);
+    invalidateCache();
```
arturbeg, line 107: makes sense, in here we want to invalidate cache if the wallet changes

src/routes/Profile/cache.ts
```diff
@@ -0,0 +1,10 @@
+import { GetProfileBadgesResponse } from '@graphql';
+
+let SEASON_BADGE_CACHE: Record<string, GetProfileBadgesResponse | undefined> = {};
```
arturbeg, line 3: do you think it's worth considering caching this data to persist even after a reload? i think that can potentially greatly improve the UX.

arturbeg (approved):

src/graphql/getters/getSeasonBadges/helpers.ts
ssbarbee: thanks I plan to add them to other utilities as I go through codebase

src/graphql/getters/getSeasonBadges/mappers.ts
ssbarbee: yes! I sat with @ioanaroceanu2000 and we did the mapping. So all you need to do in future is just extend this map

src/routes/Profile/Profile.tsx
ssbarbee: well not really... As you can see on componentDidMount (useEffect with []) and when wallet changes I invalidate the cache. So this means that whenever the user will navigate between seasons using the SeasonToggler only then we will cache and fetch from cache

src/routes/Profile/Profile.tsx
ssbarbee: not sure I understand the comment... keep in mind this code is to be removed as it mocks BE at the moment

src/routes/Profile/Profile.tsx
ssbarbee: it is a component did mount again. Maybe they went to other pages and now return to the profile page to check for badge progress

src/routes/Profile/cache.ts
ssbarbee: not really... too much caching can lead to bugs :D I thought about it but then we need to have caching per periods. Like 10min caching and I think people want data to be as live as possible. To go on traders do a trade come to profile and see their badge.

src/graphql/getters/getSeasonBadges/mappers.ts
arturbeg: amazing!

src/routes/Profile/Profile.tsx
arturbeg: sure, no worries then.

src/routes/Profile/Profile.tsx
arturbeg: got it, thanks

src/routes/Profile/Profile.tsx
arturbeg: oh i see, this essentially accounts for the scenario where they e.g. have done a trade on the trading pools page and then returned to the profile page they also see their most recent badges

arturbeg: what if instead of invalidating the entire badge cache, we edit to the list of badges?

arturbeg: also, note in order for a badge to be registered, the on-chain event data needs to be indexed by the subgraph which then feeds into the FE, however the indexing of new events takes some time and i suspect there are edgecases where users have to wait (or even refresh) the profile in order see the most recent set of their badges that accounts for their latest interactions with the protocol

src/routes/Profile/cache.ts
arturbeg: yeah, that's the ideal experience we want to achieve, there are quite a few things that feed into this e.g. time it takes for the data to propagate through the subgraph, i think we need to strike a balance with caching (combined with deliberate append/delete logic that consistently keeps the cache in sync

src/routes/Profile/Profile.tsx
ssbarbee: ```what if instead of invalidating the entire badge cache, we edit to the list of badges?``` We still have to fetch the array, the point behind cache is that we don't do API calls

src/routes/Profile/cache.ts
arturbeg: e.g. if we do a trade in the trading pool against an example position, we know for sure that we don't need to update the cache of any other position except of the position that was traded on (note there is an assumption baked into this thought, i.e. that all state changing transactions come from the UI, currently it's mostly true, though in the future we might see more transactions coming to voltz via alternative frontends/api integrators

ssbarbee: For me as enduser when I am on the page of Profile I don't want it to jump when I switch seasons all the time. But if I come from another page to the profile then it kinda makes sense to be refreshed from ground up.

What is your user experience that you are after?

src/routes/Profile/Profile.tsx
arturbeg: yeah, that's along the lines of what i'm thinking as well

arturbeg: need to edit the list, it can either increase in size or occasionally decrease in size in case we realise there is an issue with the way subgraph computes badge eligibility (thought this should not happen in theory)

src/routes/Profile/cache.ts
arturbeg: my expectation as a user is to not have a full ground up reload after returning to the page, minimises overall waiting time, though curious what others think of this UX

arturbeg: might be worth discussing with Brokoli as well

PR #320, opened 2022-11-03
fix: remove not needed inbetween components for RankingTable

ssbarbee:
remove not needed ConnectedRankingTable.tsx component
remove useRanking hook
refactor rename + move components closer to usage for better "feature" removal or experimentation

preparing it for storybook + test in upcoming PRs.


src/routes/TradingLeague/TradingLeague.tsx
```diff
@@ -1,17 +1,79 @@
...
+    }
+  };
+
+  const handleOnPrevPage = () => {
+    if (page - 1 > -1) {
+      setPage(page - 1);
+    }
+  };
+
+  const fetchRankings = async () => {
```
arturbeg, line 35: @ssbarbee might be also worth caching the rankings as well so that when i navigate back and forth from the league page, there is no need to refetch

arturbeg: though need to be mindful of the fact that the points and rank can change over time, maybe this is a stretch but could we use websockets in this context?

arturbeg (approved):

src/routes/TradingLeague/TradingLeague.tsx
ssbarbee: as you mentioned this is why caching won't work nicely here. Ideally yes websockets are nice to be used here but do we have the infra to do so? For now I think it is quite good, lets add the "live" feature if we have a "live" tournament requirment of sort :D I don't think anyone wants to sit in the leaderboard now and see how they climb

arturbeg: yeah, websockets are a stretch at this stage, in theory also could have a button in the ui that let's users manually trigger cache refresh so it doesn't automatically refresh after going back and forth from one screen to leaderboard, though again keen to discuss other ideas around how the UX should work with Brokoli's involvement as well?

PR #321, opened 2022-11-03
feat: add animation for progress bar

ssbarbee:
Animating from 0 to percentage complete! It makes the UI more lively


arturbeg (approved):
looks good, like animations:)

PR #323, opened 2022-11-04
feat: support Mellow badge

ssbarbee:
What is missing is the correct mapping to BE type. Once @ioanaroceanu2000 has this we can easliy update the FE map


ioanaroceanu2000:
let's review this after the backend integration, looks like some data is missing

src/routes/Profile/ProfilePageWalletConnected/ProfilePageWalletConnected.mocks.ts
```diff
@@ -59,6 +59,10 @@ export const season1Badges: CollectionBadge[] = [
   {
     variant: 'dryIce',
   },
+  {
+    variant: 'mellowLpVault',
+    achievedAt: new Date('02/02/2022').valueOf(),
```
ioanaroceanu2000, line 64: what if the user didn't achieve this? where do we verify this against subgraph data?

ssbarbee: this is storybook. Not used anywhere in Application. This is only for the purpose of demo. check the mappers in getSeasonBadges for the actual logic

ioanaroceanu2000: ok, got it

ioanaroceanu2000 (approved):

PR #324, opened 2022-11-04
fix: force triple equal in project via eslint

ssbarbee:
We should be using === instead of == in project. This PR adds eslint that enforces that. 
Also added more stuff to prettier check and config.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness


ioanaroceanu2000:
is this enforcing === everywhere? what if we need == at some point?

ssbarbee:
> is this enforcing === everywhere? what if we need == at some point?

Yes it enforces it everywhere. Can you give me an example why you will need to use == ever? Curios to understand this usecase.

ssbarbee:
> we often get 'any' values from the subgraph... I guess we can do casting and the use === instead

you should still write mapper for any third party data that you get. But you are missing the point... == does 2 stuff it does conversion of one type to another plus comparison. === does comparison only. If you want to compare stuff you should use === always. == is not advisable to be used because of the side effect of conversion

PR #325, opened 2022-11-04
feat/add support for minting

ssbarbee:
Added support for minting. Latest SDK was added as well.


src/graphql/getters/getSeasonBadges/getSeasonBadges.ts
```diff
@@ -43,20 +43,20 @@ export async function getSeasonBadges({
...
-    const response = SEASON_BADGE_VARIANTS[seasonId].map((badgeVariant) => {
+    return SEASON_BADGE_VARIANTS[seasonId].map((badgeVariant) => {
       const badge = badges.find((b) => BADGE_TYPE_BADGE_VARIANT_MAP[b.badgeType] === badgeVariant);
       if (!badge) {
         return {
           variant: badgeVariant as BadgeVariant,
         };
       }
       return {
+        badgeResponseRaw: badge,
```
ioanaroceanu2000, line 54: is this necessary if we already have the processed data?

ssbarbee: yes I need this so that when I consume the SDK i can provide it with the data from the graphql

src/routes/Profile/Profile.tsx
```diff
@@ -55,30 +50,73 @@ const Profile: React.FunctionComponent = () => {
...
     setCollectionBadges(result);
     setClaimButtonBulkMode('claim');
-    setClaimButtonModes(
-      SEASON_BADGE_VARIANTS[season.id].reduce(
-        (pV, cI) => ({
-          ...pV,
-          [cI as BadgeVariant]: 'claim',
-        }),
-        {} as Record<BadgeVariant, ClaimButtonProps['mode']>,
+    const claimedTodayVariants = SEASON_BADGE_VARIANTS[season.id].filter((badgeVariant) =>
```
arturbeg, line 53: why do we need to compute both claimedTodayVariants and claimedInThePastVariants ?

src/routes/Profile/Profile.tsx
```diff
@@ -55,30 +50,73 @@ const Profile: React.FunctionComponent = () => {
...
       ),
     );
+    const notClaimedVariants = SEASON_BADGE_VARIANTS[season.id].filter((badgeVariant) =>
+      result!.find(({ claimedAt, variant }) => variant === badgeVariant && !claimedAt),
+    );
+    setClaimButtonModes((p) => ({
+      ...p,
+      ...getClaimButtonModesForVariants(notClaimedVariants as BadgeVariant[], 'claim'),
+      ...getClaimButtonModesForVariants(claimedTodayVariants as BadgeVariant[], 'claimed'),
+      ...getClaimButtonModesForVariants(claimedInThePastVariants as BadgeVariant[], 'claimedDate'),
```
arturbeg, line 77: for those that are claimed in the past we want to show the claimed date for those that are today we don't?

src/routes/Profile/Profile.tsx
```diff
@@ -89,14 +127,59 @@ const Profile: React.FunctionComponent = () => {
...
+    } catch (err) {
+      setClaimButtonModes((prev) => ({
+        ...prev,
+        [variant]: 'claimError',
+      }));
+    }
   }
 
-  function handleOnClaimBulkClick(variants: BadgeVariant[]) {
+  async function handleOnClaimBulkClick(variants: BadgeVariant[]) {
```
arturbeg, line 138: this needs to be properly qa'd and also potentially worth doing some more elaborate e2e testing using this functionality as a case study to make the ui dev approach more TDD

arturbeg (approved):

src/routes/Profile/Profile.tsx
ssbarbee: due to business logic. badges claimed today should only show checkbox + claimed text. badges claimed in the past should show UI with Claimed: <DATE>. It is just for the mode

src/routes/Profile/Profile.tsx
ssbarbee: exactly! This is the UI logic I talked with Brokoli! It kinda doesnt make sense to show the date of claim if it is today.

src/routes/Profile/Profile.tsx
ssbarbee: setting e2e is not really that straight forward for the crypto world. We need to set a proper infra. But I am in the process of testing with @ioanaroceanu2000

src/routes/Profile/Profile.tsx
arturbeg: cool, makes sense

PR #326, opened 2022-11-07
feat/make collection list interactive

ssbarbee:
Trying to build connection between the list of connections and the badges overview 

https://user-images.githubusercontent.com/11079632/200266853-9a6488bf-e102-4e0f-acfb-97d565ff7c82.mov


src/routes/TradingLeague/Entry/Points/Points.tsx
```diff
@@ -30,12 +30,10 @@ export const Points: React.FunctionComponent<PointsProps> = ({ points }) => {
...
 
   return (
-    <CountUp start={0} end={points} delay={0} formattingFn={formatNumber}>
-      {({ countUpRef }) => (
-        <PointsTypography>
-          <span ref={countUpRef} />
-        </PointsTypography>
-      )}
-    </CountUp>
+    <PointsTypography>
```
ioanaroceanu2000, line 33: what's the difference here?

src/routes/TradingLeague/UserSummary/Percentage/Percentage.tsx
```diff
@@ -0,0 +1,38 @@
...
+    }}
+  >
+    {children}
+  </Typography>
+);
+
+export const Percentage: React.FunctionComponent<PercentageProps> = React.memo(({ percentage }) => {
+  return (
+    <PercentageTypography>
+      <CountUp
```
ioanaroceanu2000, line 27: Why do we need a count-up for the progress bar percentage? aren't we overcomplicating things? it seems like this value is loaded instantly.

src/routes/TradingLeague/Entry/Points/Points.tsx
ssbarbee: PointsTypography should be outer wrapper around CountUp

src/routes/TradingLeague/UserSummary/Percentage/Percentage.tsx
ssbarbee: it is nice UI touch. Both progress bar and percentage numbers will animate, giving a nice UX feel. Try the build locally. This has nothing to do with value being loaded instantly. It is just UX touch to make the UI feel more playful

ioanaroceanu2000: I thought this is the actual number i.e. 40% and it was strange to count up to only 40 (or max 100). This loading bar instead looks nice.

arturbeg:
@ssbarbee seems like there is conflict in here presumable following the merge of redeem logic to develop

src/components/interface/PortfolioHeader/PortfolioHeader.tsx
```diff
@@ -34,21 +19,11 @@ const PortfolioHeader = ({
...
-          {!isUndefined(portfolioData.totalNotional) && (
-            <>
-              {currencySymbol}
-              {formatCurrency(portfolioData.totalNotional)} {currencyCode}
-            </>
-          )}
-        </Typography>
-      </Box>
-
+      <NetNotional
```
arturbeg, line 22: nice, good idea to move it into a separate component

src/routes/Profile/BadgeCard/BadgeCard.tsx
```diff
@@ -27,53 +29,70 @@ export type BadgeCardProps = {
...
-        <ClaimButtonBox>
+export type BadgeCardHandle = {
+  scrollIntoView: () => void;
+};
+
+export const BadgeCard = forwardRef<BadgeCardHandle, BadgeCardProps>(
+  ({ onClaimButtonClick, claimButtonMode, loading, variant, claimedAt, disableClaiming }, ref) => {
+    const containerRef = useRef<HTMLElement>(null);
+    const [highlight, setHighlight] = useState(false);
+    useImperativeHandle(ref, () => ({
```
arturbeg, line 40: what's the use-case of this hook?

src/routes/Profile/BadgeCard/BadgeCard.tsx
```diff
@@ -27,53 +29,70 @@ export type BadgeCardProps = {
...
-      </DescriptionBox>
-      {disableClaiming ? null : (
-        <ClaimButtonBox>
+export type BadgeCardHandle = {
+  scrollIntoView: () => void;
+};
+
+export const BadgeCard = forwardRef<BadgeCardHandle, BadgeCardProps>(
+  ({ onClaimButtonClick, claimButtonMode, loading, variant, claimedAt, disableClaiming }, ref) => {
+    const containerRef = useRef<HTMLElement>(null);
```
arturbeg, line 38: can you explain how this reference works and what's it's utility

src/routes/Profile/ProfilePageWalletConnected/ProfilePageWalletConnected.tsx
```diff
@@ -79,7 +81,9 @@ export const ProfilePageWalletConnected: React.FunctionComponent<ProfilePageWall
 
     const collection = achievedBadgesMemo.filter((aB) => aB.achievedAt);
     const notClaimedBadges = collection.filter((b) => !b.claimedAt);
-
+    const handleSmoothScroll = React.useCallback((variant: BadgeVariant) => {
```
arturbeg, line 84: what's the visual difference of a normal vs. smooth scroll?

src/routes/TradingLeague/Entry/Points/Points.tsx
```diff
@@ -30,12 +30,10 @@ export const Points: React.FunctionComponent<PointsProps> = ({ points }) => {
...
+      <CountUp start={0} end={points} delay={0} formattingFn={formatNumber}>
```
arturbeg, line 34: btw @ioanaroceanu2000 (unrelated to this PR, but reminded me of the points issue) i think there were some users that mentioned that points have appeared to go down over time which feels strange, my current guess is that the points are not calculated using the historical exchange rates of eth at the time that points are generated and rather use the most recent exchange rate

ioanaroceanu2000: @arturbeg oh, not good, I will look into it

src/components/interface/PortfolioHeader/PortfolioHeader.tsx
ssbarbee: yep yep more control if it is smaller segment ;)

src/routes/Profile/BadgeCard/BadgeCard.tsx
ssbarbee: useImperativeHandle is a hook from React. It allows you to have a component and expose an external API for other components to use.

For example: 

```
const Parent = () => {
   const ref = useRef();
   return (
     <>
        <Child ref={ref} />
        <button onClick={() => ref.current.sayHello()} />
     </>
   )
}

const Child = React.forwardRef((props, ref) => {
   useImperativeHandle(() => ({ sayHello: alert('hello') });
   return <div>Child</div>;
});
```

We define 2 components, Child and Parent. The Child exposes API for the parent to use in any way possible. Hope this makes sense. If not I can go a bit deeper.

In the current use case, I made BadgeCard to expose and API called highlight so that when someone wants to highlight the BadgeCard then can do that by calling ref.highlight();

src/routes/Profile/BadgeCard/BadgeCard.tsx
ssbarbee: We need a ref to the div element so that you can later on use it like: containerRef.current?.scrollIntoView. Basically when we want the scrollbar to position it self on top of div.

src/routes/TradingLeague/Entry/Points/Points.tsx
ssbarbee: Seems unrelated to this :D

src/routes/Profile/BadgeCard/BadgeCard.tsx
arturbeg: i see, so the parent can reference a given child in its logic via the reference, the imperative handle is necessary for the ref constant in the parent component to work, is that correct interpretation?

src/routes/Profile/BadgeCard/BadgeCard.tsx
arturbeg: got it, thanks

src/routes/TradingLeague/Entry/Points/Points.tsx
arturbeg: yea:)

arturbeg: thanks @ioanaroceanu2000, let's create a ticket and migrate this conversation to the qa channel on discord

src/routes/Profile/ProfilePageWalletConnected/ProfilePageWalletConnected.tsx
arturbeg: noice, thanks for the pen

PR #330, opened 2022-11-07
chore/cleanup code

ssbarbee:
Minor stuff I was noticing. Remove unused code, refactor move some of the files to be closer to their Parent components.


arturbeg:
@ssbarbee looks like the prettier check is failing

config/webpack.config.js
```diff
@@ -56,9 +56,6 @@ const imageInlineSizeLimit = parseInt(process.env.IMAGE_INLINE_SIZE_LIMIT || '10
 // Check if TypeScript is setup
 const useTypeScript = fs.existsSync(paths.appTsConfig);
 
-// Check if Tailwind config exists
-const useTailwind = fs.existsSync(path.join(paths.appPath, 'tailwind.config.js'));
```
arturbeg, line 60: we don't need tailwind right?

arturbeg: what's a typical use-case for tailwind?

src/components/atomic/Background/Neons/Neons.styled.tsx
```diff
@@ -0,0 +1,15 @@
+import { styled } from '@mui/material/styles';
+import { ReactComponent as NeonBox1Svg } from './neon1.svg';
+import { ReactComponent as NeonBox2Svg } from './neon2.svg';
+
+export const NeonBox1 = styled(NeonBox1Svg)`
```
arturbeg, line 5: what if we name them "NeonBoxInner" and "NeonBoxOuter" for readability?

src/components/atomic/Loading/Loading.tsx
```diff
@@ -9,7 +9,7 @@ type LoadingProps = {
 const Loading: React.FunctionComponent<LoadingProps> = ({ sx = {} }) => {
   return (
     <Box sx={{ height: 40, width: 40, ...sx }}>
-      <img src="/images/loading.gif" alt="image" height="100%" width="100%" />
+      <img src="/images/loading.gif" height="100%" width="100%" />
```
arturbeg, line 12: do we need an altprop in here as suggested by gh actions?

src/components/interface/PositionTable/components/PositionTableHead/PositionTableHead.tsx
```diff
@@ -110,7 +109,7 @@ const PositionTableHead: React.FunctionComponent<PositionTableHeadProps> = ({
             }}
           >
             <BulletLabel
-              sx={{ color: getFixedRateHealthTextColor(fixedRateHealthFactor) }}
+              sx={{ color: getHealthTextColor(fixedRateHealthFactor) }}
```
arturbeg, line 112: it's important to note that there are two different types of health colours:
1. health colour of a position which is a function of their margin requirement (tells a user how safe their position is from being liquidated)
2. health colour (specific to liquidity providers), this one e.g. turns red when the current fixed rate in the amm is within the fixed rate range of the liquidity provider, yellow when it's close to a boundary of the range and green when it's within the range and not close to the boundaries (this one only applies to the positions in the lp positions screen)

arturbeg: think it's important to make sure these are distinguishable in the code

yarn.lock
```diff
@@ -7747,16 +7747,7 @@ acorn-jsx@^5.3.1, acorn-jsx@^5.3.2:
...
-  version "1.8.2"
-  resolved "https://registry.npmjs.org/acorn-node/-/acorn-node-1.8.2.tgz"
-  integrity sha512-8mt+fslDufLYntIoPAaIMUe/lrbrehIiwmR3t2k9LljIzoigEPF27eLk2hy8zSGzmR/ogr7zbRKINMo1u0yh5A==
-  dependencies:
-    acorn "^7.0.0"
-    acorn-walk "^7.0.0"
-    xtend "^4.0.2"
-
-acorn-walk@^7.0.0, acorn-walk@^7.1.1, acorn-walk@^7.2.0:
+acorn-walk@^7.1.1, acorn-walk@^7.2.0:
```
arturbeg, line 7750: is there another conflict with the sdk dependencies similar to what we saw with the apollo client for the graph?

config/webpack.config.js
ssbarbee: tailwindcss is just a toolkit you use when you want to get some fancy classes in CSS that help you build nice UI. Since we use Material UI there is no need for anything like this. Previous FE people seemed to copy pasted some ultra generic webpack config that just slow us down with the extra "crap" in it.

src/components/atomic/Loading/Loading.tsx
ssbarbee: I think gh actions commented that we don't need alt for this case.

src/components/atomic/Background/Neons/Neons.styled.tsx
ssbarbee: sure can work better!

src/components/interface/PositionTable/components/PositionTableHead/PositionTableHead.tsx
ssbarbee: I kinda disagree to have same function with same logic just because it is used in different places. Think for example parseInt, you use it everywhere yet we don't call it parseIntForXFeature, parseIntForYFeature.

Perhaps we can think of better naming for getHealthTextColor that tries to unify both ideas in one?

Open for suggestions from your end!

ssbarbee:
> @ssbarbee looks like the prettier check is failing

fixed

src/components/atomic/Background/Neons/Neons.styled.tsx
ssbarbee: fixed in: 45b92be513a92659393013afe709b0706e64af8e

src/components/atomic/Loading/Loading.tsx
ssbarbee: fixed in: 8fe2861f0d791155669d7d1c7cb0988318af25c8

PR #332, opened 2022-11-07
fix: add back phase 1 query

ioanaroceanu2000:
added back the getter with the phase 1 Subgraph query 
Profilev1 now points to this getter and is controlled by REACT_APP_COMMUNITY_P2=UNPROVIDED


ssbarbee (approved):
LGTM

PR #337, opened 2022-11-08
fix: token approval for fixed borrow product

avniculae:


src/contexts/BorrowFormContext/BorrowFormContext.tsx
```diff
@@ -118,10 +118,19 @@ export const BorrowFormProvider: React.FunctionComponent<BorrowFormProviderProps
     validateId.current = (validateId.current + 1) % 666013;
     const currentValidateId = validateId.current;
 
+    updateBorrowSwapInfoAsyncState();
```
ioanaroceanu2000, line 121: do we want this to be called at every rerender?

src/contexts/BorrowFormContext/BorrowFormContext.tsx
```diff
@@ -356,8 +367,21 @@ export const BorrowFormProvider: React.FunctionComponent<BorrowFormProviderProps
         });
       }
     }
+  };
+
+  useEffect(() => {
+    void selectedFixedDebtChanged();
```
ioanaroceanu2000, line 373: why is this called using void?

avniculae: saw it in other places in the file and thought we want to stick to this. However I'm not sure who introduced this in the first place anx whether we do want to stick to this. @ssbarbee any thoughts on this?

ssbarbee: The void is used when you don't want to await an async function. In this context it doesn't make sense. I think you can remove the void. 

It resolves no-floating-promises because it does something (or rather, explicitly nothing) with the promise.

src/contexts/BorrowFormContext/BorrowFormContext.tsx
```diff
@@ -356,8 +367,21 @@ export const BorrowFormProvider: React.FunctionComponent<BorrowFormProviderProps
...
   }, [selectedFixedDebt]);
 
+  useEffect(() => {
+    if (!approvalsNeeded && touched.current.length) {
+      void selectedFixedDebtChanged();
+      if (selectedFixedDebt === 0) {
+        void validate();
```
ssbarbee, line 380:
```suggestion
       validate();
```

src/contexts/BorrowFormContext/BorrowFormContext.tsx
```diff
@@ -356,8 +367,21 @@ export const BorrowFormProvider: React.FunctionComponent<BorrowFormProviderProps
...
     }
+  };
+
+  useEffect(() => {
+    void selectedFixedDebtChanged();
   }, [selectedFixedDebt]);
 
+  useEffect(() => {
+    if (!approvalsNeeded && touched.current.length) {
+      void selectedFixedDebtChanged();
```
ssbarbee, line 378:
```suggestion
     selectedFixedDebtChanged();
```

ssbarbee (approved):

PR #342, opened 2022-11-09
fix/improvements around claiming

ssbarbee:
* prevent claiming when claiming in progress (example: claiming in progress for single button, but bulk claim is being pressed).
* prevent season change when claiming in progress


ioanaroceanu2000:
lgtm

PR #344, opened 2022-11-10
fix/claiming qa bugbash fixes

ssbarbee:
Fixed:
* season change loses claim status
* after claim the number doesn't refresh

### DIFF
--- a/src/routes/Profile/Profile.tsx
+++ b/src/routes/Profile/Profile.tsx
@@ -125,12 +125,13 @@ const Profile: React.FunctionComponent = () => {
         ...prev,
         [variant]: 'claimed',
       }));
-      setCollectionBadges((prev) => {
-        return prev.map((b) => ({
-          ...b,
-          claimedAt: b.variant === variant ? Date.now().valueOf() : b.claimedAt,
-        }));
-      });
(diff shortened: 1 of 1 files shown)


src/routes/Profile/Profile.tsx
```diff
@@ -125,12 +125,13 @@ const Profile: React.FunctionComponent = () => {
...
       }));
-      setCollectionBadges((prev) => {
-        return prev.map((b) => ({
-          ...b,
-          claimedAt: b.variant === variant ? Date.now().valueOf() : b.claimedAt,
-        }));
-      });
+      const nextCollectionBadges = collectionBadges.map((b) => ({
+        ...b,
+        claimedAt: b.variant === variant ? Date.now().valueOf() : b.claimedAt,
```
ioanaroceanu2000, line 130: what is this `nextCollectionBadges` used for? is this parsing through all awarded badges and setting `claimedAt` either at `now` or `time of claiming`?

ssbarbee: `nextCollectionBadges` is just temporary variable in this context, used to determine what is the next state for the collection badges. Since I don't want to "requery" the subgraph once again I edit the already in-state array.

PR #345, opened 2022-11-10
chore/release 10112022

ssbarbee:


ssbarbee:
> can we add a description to the release: what changes does it contain? This wouldn't be necessary if we didn't have this long list of commits. Is there a way to only show release-specific commits?

We can add if we had a way of linking each branch with ticket. Right now we dont have effective ticket system, so this is a bit hard to follow. All commits are release specific.

ioanaroceanu2000:
> > can we add a description to the release: what changes does it contain? This wouldn't be necessary if we didn't have this long list of commits. Is there a way to only show release-specific commits?
> 
> We can add if we had a way of linking each branch with ticket. Right now we dont have effective ticket system, so this is a bit hard to follow. All commits are release specific.

1) I can see >150 commits from 17 days ago. are you sure these commits are not yet in main? The changes don't seem to reflect that many commits. 2) can we make https://chore-release-10112022.d2f2fhja7tsifl.amplifyapp.com/ work pls? to check if there are no major bugs

ssbarbee:
> > > can we add a description to the release: what changes does it contain? This wouldn't be necessary if we didn't have this long list of commits. Is there a way to only show release-specific commits?
> > 
> > 
> > We can add if we had a way of linking each branch with ticket. Right now we dont have effective ticket system, so this is a bit hard to follow. All commits are release specific.
> 
> 1. I can see >150 commits from 17 days ago. are you sure these commits are not yet in main? The changes don't seem to reflect that many commits. 2) can we make https://chore-release-10112022.d2f2fhja7tsifl.amplifyapp.com/ work pls? to check if there are no major bugs

yes the main problem is that we squash merge... and those commits are not on main, need to rethink this strategy thanks!

PR #351, opened 2022-11-11
fix: max withdrawal bugs

avniculae:


src/components/composite/MarginAmount/MarginAmount.tsx
```diff
@@ -50,12 +50,16 @@ const MarginAmount: React.FunctionComponent<MarginAmountProps> = ({
...
 
-  const maxAmountToWithdraw =
+  let maxAmountToWithdraw =
     !isUndefined(marginAmount) && !isUndefined(initialMarginRequirement)
-      ? marginAmount - initialMarginRequirement
+      ? Math.max(marginAmount - initialMarginRequirement, 0)
       : undefined;
+  if (maxAmountToWithdraw) {
+    maxAmountToWithdraw = Math.floor(maxAmountToWithdraw * 10000) / 10000;
+  }
```
ssbarbee, line 59: you can merge this Math.max and Math.floor in one line and make maxAmountToWithdraw a const. WDYT?

src/hooks/useCurrentPositionMarginRequirement.ts
```diff
@@ -1,17 +1,25 @@
...
     useState<number>();
 
   // Load the users balance of the required token
   useEffect(() => {
     const getData = async () => {
-      try {
-        const mrm = await amm.getPositionMarginRequirement(1, 999);
-        setCurrentPositionMarginRequirement(mrm);
-      } catch (e) {
+      if (fixedRateLower && fixedRateUpper) {
```
ssbarbee, line 15: is the check isUndefined missing here?

avniculae: doesn't it have the same effect?

ssbarbee: currently it also protects you against 0s, nulls, empty strings, not sure if this is desired

undefined && undefined === false
0 && 0 === false
null && null === false
'' && '' === false

ssbarbee (approved):

PR #359, opened 2022-11-14
refactor/remove not needed deps

ssbarbee:
[@ioanaroceanu2000] https://refactor-remove-not-needed-deps.d2f2fhja7tsifl.amplifyapp.com/
* remove not needed deps, some leftovers from CRA eject
* add semantic release for automatically bumping FE version


package.json
```diff
@@ -49,11 +49,18 @@
     "plugins": [
       "@semantic-release/commit-analyzer",
       "@semantic-release/release-notes-generator",
+      [
+        "@semantic-release/changelog",
```
Sol3iman, line 53: Perhaps, I am not fully filled in. This package allows us to automatically update the package versions of our node project or does it mean something else? @ssbarbee

ssbarbee: hey @Sol3iman yes! this is quite popular package: https://www.npmjs.com/package/@semantic-release/npm. It is used to automatically bump the package.json version based on commit messages since last bump. Each commit message should follow the https://www.conventionalcommits.org/en/v1.0.0/ standard (which we do) and based on those a MAJOR, MINOR or PATCH bump is done.

This can be used for any package really. We set it up with @0xZenus for the SDK V2 as well. All it needs is permissions to push to the repository so it can automatically bump it. 

Also for this package there are plenty of plugins such as automatic changelog updates and so on. Let me know if this helps

Sol3iman: I really like this and the links are super useful. Thank you @ssbarbee . Can you quickly elaborate a basic use-case flow? For example, I have decided to update the ethers dependency to the latest version. Does semantics pick this up and update the version automatically?

ioanaroceanu2000 (approved):

package.json
ssbarbee: Hey @Sol3iman.

Like I mentioned semantic release analyses the commit messages since last release.

For example, lets say current package.json version is 0.0.1.

Example 1:
If you merge a pull request that has commit messages:

fix: bump depa
chore: fix bug 123

Semantic release will analyse them and weigh each one. In this case the strongest is "fix" and it corresponds to PATCH bump, meaning it updates the next version to 0.0.2.


Example 2: 
If you merge a pull request that has commit messages:

feat: next big thing
fix: fix bug 123

Semantic release will analyse them and weigh each one. In this case the strongest is "feat" and it corresponds to MINOR bump, meaning it updates the next version to 0.1.0.

Example 3:
If you merge a pull request that has commit messages:

fix: bump depa
BREAKING CHANGE: api restructed

Semantic release will analyse them and weigh each one. In this case the strongest is "BREAKING CHANGE" and it corresponds to MAJOR bump, meaning it updates the next version to 1.0.0.


Hope this makes sense! 🙏

Sol3iman: Perfect, this makes sense. Thank you

Sol3iman (approved):

PR #372, opened 2022-11-17
Config file for mellow

0xZenus:


src/config.ts
```diff
@@ -0,0 +1,55 @@
+import { providers } from "ethers";
```
ssbarbee, line 1: @0xZenus bring this closer to usage. Place under hooks/useMellowLPVaults directory along side your hook.

0xZenus: I was thinking to treat it as a global configuration file in the future. Maybe we can process all env variable in here at once. What do you think? There's a lot to off-load (factory address, network names, ids, whitelisted addresses etc.)

ssbarbee: sounds good but until we do that lets move it closer to usage, since it is easier to delete :) You can always move it away once you start using it across project

0xZenus: agree.

0xZenus: https://github.com/Voltz-Protocol/voltz-ui/pull/372/commits/688bc5da2bec4d30c58dc1afb2d4401b10fc4e2b

src/hooks/useMellowLPVaults/config.ts
```diff
@@ -0,0 +1,61 @@
...
+  MELLOW_VAULTS: {
+    voltzVault: string;
+    erc20RootVault: string;
+    erc20RootVaultGovernance: string;
+  }[];
+  PROVIDER: providers.BaseProvider;
+} => {
+  const network = process.env.REACT_APP_NETWORK;
+
+  if (isUndefined(network)) {
```
ssbarbee, line 42: isUndefined from lodash gives false when input is null.

perhaps use `!network` for this place. I am not sure how the process.env.VARIABLE can be passed. Maybe it can be null somehow :D

0xZenus: sure. actually, I have one question here, when network is not included in the config, do we want to throw error or return some kind of default value?

ssbarbee: ideally we want to have sentry error reported to us... I think that these variables must be passed since our App cannot work otherwise. So I would say add a TODO: next to your throw new Error and we handle it better once our project setup allows.

ssbarbee: for me this is where context + context variables really shine. We can have <ENVsProvider /> component that can resolve this for us. And we can have a hook useEnvVars() that we use to read from Context in any component basically.

0xZenus: Yes, we let's create a task around this in the BAU section where we also mention off-loading other env. variables.

0xZenus: sure. tests are failing at this moment because network is not specific. where do we set env. variables for the tests?

ssbarbee: hmmmm we only have tests for Profile... Why don't you call getConfig() where you needed instead of calling it when file is bundled?

0xZenus: ok. I can do it.

ssbarbee (approved):

PR #374, opened 2022-11-17
feat: add support for copying a referrer link

ssbarbee:
* consume heroku database to copy the referrer link
* testing url: https://feat-consume-database-to-generate-invite-link.d2f2fhja7tsifl.amplifyapp.com/#/trader-pools


src/routes/Profile/get-referrer-link.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (!data) {
+      return undefined;
+    }
+    const status = data.status ? parseInt(data.status, 10) : 200;
+    if (status !== 200 || !data.refers_with_code) {
+      return undefined;
+    }
+    const link = `${window.location.origin}/?invitedBy=${data.refers_with_code}`;
+    debugger;
```
CyclopsRex: delete?

src/routes/Profile/get-referrer-link.ts
```diff
@@ -0,0 +1,32 @@
+import axios from 'axios';
+
+export const getReferrerLink = async (account: string) => {
+  const baseUrl = process.env.REACT_APP_REFERRAL_AND_SIGNATURE_SERVICE_URL;
```
CyclopsRex, line 4: As discussed I am defaulting this in https://github.com/Voltz-Protocol/voltz-ui/pull/375 if not set, cos that seemed better than errors or weird behaviour that we might not catch in testing. Totally open to not defaulting but we should probably be consistent.

src/routes/Profile/get-referrer-link.ts
```diff
@@ -0,0 +1,32 @@
...
+        description?: string;
+        refers_with_code?: string;
+        wallet_address?: string;
+      }>(`${baseUrl}/get-refers-with/${account}`)
+    )?.data;
+
+    if (!data) {
+      return undefined;
+    }
+    const status = data.status ? parseInt(data.status, 10) : 200;
```
CyclopsRex, line 22: I fixed the API to return 404 properly if not found so if you wrote/tested this code before that then the error handling here may need to change.

src/routes/Profile/Profile.tsx
```diff
@@ -205,20 +207,27 @@ const Profile: React.FunctionComponent = () => {
...
   }
 
-  function handleOnCopyLinkButtonClick() {
-    // todo: implement logic for fetching the referrer link
+  async function handleOnCopyLinkButtonClick() {
+    if (!wallet.account) {
+      return;
+    }
     setCopyLinkButtonMode('copying');
+    const link = await getReferrerLink(wallet.account);
```
CyclopsRex, line 215: Do I understand correctly that this only gets called when they click the link? Could take a few hundred ms to get code from API in the worst case, so is there any value in pre-fetching the code when we load the profile page?

src/routes/Profile/get-referrer-link.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (status !== 200 || !data.refers_with_code) {
+      return undefined;
+    }
+    const link = `${window.location.origin}/?invitedBy=${data.refers_with_code}`;
```
CyclopsRex: A quick test suggests that this doesn't work with https://github.com/Voltz-Protocol/voltz-ui/pull/375, which was tested with URIs of the form http://localhost:3000/#/trader-pools?invitedBy=SWAEUlit cos that's what we used in the hackathon. If you think this URL without the `/#/trader-pools` part is better, then please can you suggest a new home for the localStorage-writing code in https://github.com/Voltz-Protocol/voltz-ui/pull/375 ?

CyclopsRex (commented):
Some comments below. Wonder if we should merge this with https://github.com/Voltz-Protocol/voltz-ui/pull/375 and test them together before merging into develop. But I'm out tomorrow so I'll leave this up to you.

src/routes/Profile/get-referrer-link.ts
ssbarbee: fixed in 5e7aa349877b2a2309a6a6e6afbe1474d8740b48

src/routes/Profile/get-referrer-link.ts
ssbarbee: Me and @0xZenus created ticket on this matter. We will have a look into this in separate ticket addressing most of ENVs

src/routes/Profile/Profile.tsx
ssbarbee: I don't want to "create" network calls if not needed. Create them per request of user. Potentially I can add caching if the same "code" is returned always

src/routes/Profile/get-referrer-link.ts
ssbarbee: yes I will fix this in #375 and test it accordingly. Thanks for pointing it out.

CyclopsRex: Great, thanks! Please resolve this conversation once fixed in #375

src/routes/Profile/Profile.tsx
CyclopsRex: Same code is always returned for a given user.

0xZenus (approved):

PR #375, opened 2022-11-17
feat: referral tracking

CyclopsRex:
Track incoming referrals from the URL parameters, e.g. `http://localhost:3000/#/trader-pools?invitedBy=SWAEUlit`, by adding them to a database via the `/add-signature` API endpoint.

Piggybacking on the `/add-signature` (which checks signature validity) prevents referrals being added to the database by any account other than the account being referred.

Note that the referral is completely silent from the PoV of the account being referred, and there is no longer any signature content to have them epxlicitly acknowledge the referral. In a world where referred accounts receive no pointz or other benefit for being referred, this explicit acknowledgement within the signed message seemed to add little except unnecessary complexity.

Note: only valid referral IDs (eight characters, and stored in the database) will be accepted, but others should be silently ignored

@ssbarbee a few things requiring attention here:
 - we need to merge and test with https://github.com/Voltz-Protocol/voltz-ui/pull/374
 - where is really the best place to read the URL param (invitedBy), and does it seem OK to localStorage until we connect a wallet?
 - (related to the above) where to store an `invitedBy` string constant that can be used by both the reading and writing code (if we still write to local storage)
 - we should add google analytics to monitor referral links being followed, and then track which wallets are connected using each referral link. I think you said you'd have a go at doing this and you could walk me through it on Mon?
 - @CyclopsRex to test with the existing version of the REST API to ensure that the UIs use of the API is back-compatible


ssbarbee:
> * we still need the "invite people to Voltz using this link" bit of the UI; believe you have designs are are working on this already

this is being taken care of in another PR https://github.com/Voltz-Protocol/voltz-ui/pull/374 have a look

src/routes/Trader/Trader.tsx
```diff
@@ -43,10 +43,13 @@ const Trader: React.FunctionComponent = () => {
...
   const [searchParams] = useSearchParams();
-  if (!!searchParams.get(referrerKey) && !localStorage.getItem(referrerKey)) {
+  if (
+    (!!searchParams.get(referrerKey) && !localStorage.getItem(referrerKey)) ||
+    localStorage.getItem(referrerKey)?.length !== 8 // Earlier, pre-release keys may have used more than 8 characters but we overwrite these
+  ) {
     // Referrer is set in URL and has not already been saved in storage; save to storage now
     // We do deliberately do not overwrite any existing value because we want to credit the first referral link that was followed
-    localStorage.setItem('invitedBy', searchParams.get('invitedBy') || '');
+    localStorage.setItem(referrerKey, searchParams.get(referrerKey) || '');
```
ssbarbee, line 52: should we have a function to construct the referrerKey? I think this key should depend on "env" where testing happens. What if someone is testing on test.voltz.xyz and we set stuff to localStorage. Then they go on real app.voltz.xyz and they get invite.... Because they are part of our test group the app wont for them on production.

CyclopsRex: I think localStorage is per domain and port, so values set on testnet won't be read on mainnet and vice versa. Please shout if I've misunderstood.

CyclopsRex:
> > * we still need the "invite people to Voltz using this link" bit of the UI; believe you have designs are are working on this already
> 
> this is being taken care of in another PR #374 have a look

Thx. PR description updated.

src/routes/Trader/Trader.tsx
ssbarbee: thakns for sharing this!

0xZenus (approved):

src/App.tsx
```diff
@@ -1,4 +1,4 @@
-import { Routes, Route, Navigate } from 'react-router-dom';
+import { Routes, Route, Navigate, useSearchParams } from 'react-router-dom';
```
ssbarbee, line 1: todo: GA tracking

ssbarbee (changes requested):

src/App.tsx
```diff
@@ -15,14 +15,43 @@ import {
...
+    }
+
+    const isValidReferrer = isValidReferrerValue(searchParamsReferrer);
+    if (!isValidReferrer) {
+      return;
+    }
+
+    // Referrer is set in URL and has not already been saved in storage; save to storage now
+    // We do deliberately do not overwrite any existing value because we want to credit the first referral link that was followed
+    if (!isRefererStored()) {
```
CyclopsRex, line 50: Do we need isReferrerStored or would a call to isValidReferrerStored suffice here?

src/App.tsx
```diff
@@ -15,14 +15,43 @@ import {
...
   useEffect(() => {
     if (process.env.REACT_APP_GTM_CODE) {
       TagManager.initialize({ gtmId: process.env.REACT_APP_GTM_CODE });
     }
   }, []);
 
+  // referrer logic - run everytime params change for invitedBy
+  useEffect(() => {
+    // Earlier, pre-release keys may have used more than 8 characters, but we overwrite these
+    if (!isValidReferrerStored()) {
```
CyclopsRex, line 39: Do we need this block of code or would the one on line 50 (with the suggested change to use isValidReferrerStored) be ok on its own?

CyclopsRex (commented):
Thanks for doing this! Lgtm. Added some questions but if you don't think I'm onto something feel free to merge as is.

src/App.tsx
ssbarbee: well this is something that happens first. Even if they don't have a search param (missing ?invitedBy=XXX), but have old invalid value in localstorage we will reset it. I tried to split the logic you did and make it more modular.

src/App.tsx
ssbarbee: isValidReferrerStored will work as well but isValidReferrerStored = getReferrer + isValidReferrer = 2 fn calls (this is 1).
since this happens after isValidReferrerStored which is on line 39, if there is no valid referrer in localstorage we would delete it, so this will and is more optimal imho.

src/App.tsx
ssbarbee: created ticket on this topic: https://www.notion.so/voltz/GA-analytics-for-Referrer-requirments-ca87cc78fabe443e977f2572fe2c74ab @CyclopsRex I think once we have it clear what we are after we can do the impl. right now different people have different ideas imho

src/App.tsx
CyclopsRex: Understood, and fine to keep if you want, but I'm not sure we need to do that. Sufficient to just replace it with something new (or leave it alone if there's nothing new).

src/App.tsx
CyclopsRex: Ok. If you don't make the other change then you don't need to change this. (But if you do you do. :)

src/App.tsx
ssbarbee: I will keep it as it is. To my mind cleaning up ensures we have a clean state for the "next feature" we want to build on top of this.

ssbarbee (approved):

PR #385, opened 2022-11-21
fix/update copies

ssbarbee:


ioanaroceanu2000 (approved):
One thing to keep in mind is to avoid showing these badges in seasons when they can't be achieved. Anyways, LGTM for now

PR #389, opened 2022-11-21
chore: allow non-empty values treated as false for REACT_APP_SKIP_TOS_CHECK

CyclopsRex:
@ssbarbee per discord discussion - open to a better approach if there is one


src/contexts/WalletContext/ProviderWrapper.tsx
```diff
@@ -70,7 +70,10 @@ const ProviderWrapper: React.FunctionComponent<ProviderWrapperProps> = ({
           const walletAddress = await newSigner.getAddress();
 
           // Do checks that could stop us allowing the wallet to connect
-          if (!process.env.REACT_APP_SKIP_TOS_CHECK) {
+          if (
+            !process.env.REACT_APP_SKIP_TOS_CHECK ||
```
ssbarbee, line 74: @CyclopsRex throughout the project we use the value of "UNPROVIDED" when we want to say an ENV variable is present but ignore it's value. Is it possible to use the same idea here?

I would also write the code in better fashion.

```
const skipTOSCheck = process.env.REACT_APP_SKIP_TOS_CHECK && process.env.REACT_APP_SKIP_TOS_CHECK !== 'UNPROVIDED';

if(!skipTOSCheck) { await services.checkForTOSSignature(newSigner, walletAddress); }
```

WDYT? I don't like to introduce another set of ideas when ENV variable is not present, like 'n', 'no', 'false'.

CyclopsRex: @ssbarbee I think the string "UNPROVIDED" is a weird way to express the semantics false for a boolean, and if we do that for other _boolean_ environment variables (I haven't checked) then I think we should consider changing it.

But I don't care enough to spend more time discussing, so I've gone ahead and made your suggested change. :)

ssbarbee: thanks! Ideally we should have one util where we check this. Now it is spread across codebase so it is wild west in terms of checking if "ENV" var is present or not. I do agree that "UNPROVIDED" is weird, but it is what is used

ssbarbee (approved):

PR #401, opened 2022-11-24
feat: integrate with sdk ranking

ioanaroceanu2000:


src/routes/TradingLeague/TradingLeague.tsx
```diff
@@ -32,26 +34,24 @@ const TradingLeague: React.FunctionComponent = () => {
...
     setLoading(false);
   };
 
   const setUserRanking = (rankingResults: RankType[], walletAddress?: string | null) => {
-    const userEntry: RankType | undefined = rankingResults.find((r) => r.address === walletAddress);
-
+    const userEntry: RankType | undefined = rankingResults.find((r) => r.address.toLowerCase() === walletAddress?.toLowerCase());
     setUserPoints(userEntry?.points || -1);
-    setUserRank(userEntry?.rank || -1);
+    setUserRank( userEntry ? (isUndefined(userEntry.rank) ? -1 : userEntry.rank) : -1);
```
ssbarbee, line 54: how is this different to ```    setUserRank(userEntry?.rank || -1);```

src/routes/TradingLeague/Leaderboard/Leaderboard.tsx
```diff
@@ -9,7 +9,7 @@ import ChevronLeft from '@mui/icons-material/ChevronLeft';
 import { Header } from '../Header/Header';
 import { Entry } from '../Entry/Entry';
 import { LeaderboardHeader } from '../LeaderboardHeader/LeaderboardHeader';
-import { RankType } from '../../../graphql';
+import { RankType } from '@voltz-protocol/v1-sdk/dist/types/entities';
```
ssbarbee, line 12: export rank type properly please. I want use to import from '@voltz-protocol/v1-sdk'

src/routes/Profile/helpers.ts
```diff
@@ -208,3 +209,32 @@ export const getClaimButtonModesForVariants = (
...
   `${userId.toLowerCase()}#${seasonId}`;
+
+export const getSDKInitParams = (signer: Signer | null) => {
+    const ignoredWalletIds = process.env.REACT_APP_IGNORED_LEAGUE_WALLETS &&
+      process.env.REACT_APP_IGNORED_LEAGUE_WALLETS !== 'UNPROVIDED'
+        ? process.env.REACT_APP_IGNORED_LEAGUE_WALLETS.split(',')
+            .map((s) => s.trim().toLowerCase())
+            .reduce((pV, cI) => ({ ...pV, [cI]: true }), {})
+        : {};
+    if (!process.env.REACT_APP_COMMUNITY_SBT_ADDRESS ||
```
ssbarbee, line 220: move this on top of function. No need to do the complex reduce if this if fails

src/routes/Profile/data/getSeasonBadges/getSeasonBadges.ts
```diff
@@ -2,40 +2,30 @@ import { Season } from '../../../../hooks/season/types';
...
+  sbt
 }: {
   userId: string;
-  seasonId: Season['id'];
-  signer: Signer;
+  season: Season;
+  sbt: CommunitySBT
 }): Promise<GetProfileBadgesResponse> {
   if (!process.env.REACT_APP_SUBGRAPH_BADGES_URL) {
-    return getDefaultResponse(seasonId);
```
ssbarbee, line 18: next time around you can just do `const seasonId = season.id` and don't have to change 5 lines below :)

src/routes/Profile/Profile.tsx
```diff
@@ -37,11 +37,15 @@ const Profile: React.FunctionComponent = () => {
...
+    const params = getSDKInitParams(wallet.signer);
+    if (!params) {
+      return getDefaultResponse(season.id);
+    }
     return await getSeasonBadges({
       userId: account,
-      seasonId: seasonId,
-      signer: wallet.signer as Signer,
+      season: season,
+      sbt: new CommunitySBT(params)
```
ssbarbee, line 48: why don't we construct the sbt at the `getSeasonBadges` utility. This way UI keeps UI stuff and 0 knowledge of SDK for this purpose

src/routes/TradingLeague/TradingLeague.tsx
ioanaroceanu2000: userEntry.rank can be undefined even if is defined

src/routes/TradingLeague/TradingLeague.tsx
```diff
@@ -32,26 +34,24 @@ const TradingLeague: React.FunctionComponent = () => {
...
-      ignoredWalletIds:
-        process.env.REACT_APP_IGNORED_LEAGUE_WALLETS &&
-        process.env.REACT_APP_IGNORED_LEAGUE_WALLETS !== 'UNPROVIDED'
-          ? process.env.REACT_APP_IGNORED_LEAGUE_WALLETS.split(',')
-              .map((s) => s.trim().toLowerCase())
-              .reduce((pV, cI) => ({ ...pV, [cI]: true }), {})
-          : {},
-    });
+    let result: RankType[] = [];
+    const params = getSDKInitParams(signer);
```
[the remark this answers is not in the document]
ioanaroceanu2000: I removed the return case. leaving the env vars check for the SDK ce1158a161687e6b84a00e7a94627b53bfe1607c

src/routes/Profile/helpers.ts
ioanaroceanu2000: I removed this check, SDK will return the default answer is params are missing ce1158a161687e6b84a00e7a94627b53bfe1607c

src/routes/Profile/Profile.tsx
ioanaroceanu2000: I added a utility to give us Community SBT directly without the params ce1158a161687e6b84a00e7a94627b53bfe1607c. Does this sound ok?

src/routes/Profile/Profile.tsx
```diff
@@ -37,11 +37,15 @@ const Profile: React.FunctionComponent = () => {
 
   const [copyLinkButtonMode, setCopyLinkButtonMode] = useState<CopyLinkButtonProps['mode']>('copy');
 
-  const fetchBadges = async (seasonId: Season['id'], account: string) => {
+  const fetchBadges = async ( account: string) => {
+    const params = getSDKInitParams(wallet.signer);
+    if (!params) {
+      return getDefaultResponse(season.id);
```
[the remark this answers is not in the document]
ioanaroceanu2000: I removed this statement. Not the SDK handles it ce1158a161687e6b84a00e7a94627b53bfe1607c

src/routes/TradingLeague/TradingLeague.tsx
ssbarbee: but this code; setUserRank(userEntry?.rank || -1); already takes care of that. If undefined I default it to -1. not sure why we need to complicate it with isUndefined logic

src/routes/Profile/Profile.tsx
ssbarbee: sure works for now! thanks!

src/routes/TradingLeague/TradingLeague.tsx
ioanaroceanu2000: oh, my bad. so the issue here is that if rank === 0 then it returns -1 but it should actually be #1. not sure what the logic behind this is but the code above seems to fix the problem of the user's profile showing rank -- when they are actually first.

src/routes/TradingLeague/Leaderboard/Leaderboard.tsx
ioanaroceanu2000: fixed ineff79d36

src/routes/TradingLeague/TradingLeague.tsx
ssbarbee: then setUserRank(userEntry?.rank ?? -1) is much nicer and cleaner way to solve it. Nice catch for the bug

ioanaroceanu2000: 7499fabd replaced with proposed `userEntry?.rank ?? -1`

ssbarbee (approved):

PR #404, opened 2022-11-25
refactor/move amms to higher level

ssbarbee:


0xZenus (commented):
When doing this, we need to be very careful at filtering the amms because it happens to show some pools to LPs and some to Traders.

0xZenus (approved):

PR #410, opened 2022-11-28
fix: remove abs for cashflow

ssbarbee:


src/components/interface/PositionTable/components/PositionTableRow/components/CurrentMargin/CurrentMargin.tsx
```diff
@@ -43,7 +43,7 @@ export const CurrentMargin: React.FunctionComponent<CurrentMarginProps> = ({
       return 'Margin';
     }
 
-    const formattedAccFlow = formatCurrency(Math.abs(accruedCashflow));
```
0xZenus, line 46: We need to undo this change and check against ``accruedCashflow`` in here ``{accFlowValue > 0 ? '+' : accFlowValue < 0 ? '-' : ''}`` (instead of ``accFlowValue``).

0xZenus (changes requested):

src/components/interface/PositionTable/components/PositionTableRow/components/CurrentMargin/CurrentMargin.tsx
ssbarbee: fixed in c6182399cab3cc2cd6d3f4eb3a493b035d3a9344 have a look

0xZenus (approved):

PR #421, opened 2022-11-30
feat: add sentry tracking

ssbarbee:


src/contexts/MintBurnFormContext/MintBurnFormContext.tsx
```diff
@@ -436,20 +436,16 @@ export const MintBurnFormProvider: React.FunctionComponent<MintBurnFormProviderP
...
-          }
+      const hasEnoughFunds = await hasEnoughUnderlyingTokens(
+        positionAmm || poolAmm,
+        margin,
+        mode === MintBurnFormModes.ROLLOVER ? position : undefined,
+      );
+      if (!hasEnoughFunds) {
+        valid = false;
+        if (touched.current.includes('margin')) {
+          err['margin'] = 'Insufficient funds';
```
Sol3iman, line 447: Has the catch block here been removed in favour of the Sentry error model?

ssbarbee: no not really @Sol3iman if you have a closer look it seems that hasEnoughUnderlyingTokens already has a try catch... So having try catch here as well is redundant since the catch will never be executed. Kinda like dead code.

Sol3iman: Oh I see, the if clause handles that. Interesting. In that case this PR LGTM

Sol3iman (approved):

PR #422, opened 2022-11-30
release/30112022 sentry tracking

ssbarbee:
testing url: https://release-30112022-sentry-tracking.d1zp7m1trbs6ja.amplifyapp.com/#/trader-pools


scripts/aws/README.md
```diff
@@ -0,0 +1,30 @@
+# Why?
```
Sol3iman, line 1: So `@aws-sdk/client-s3 has to be installed globally on a machine before submitting a new PR with commit msg feat/ so that the bucket is deleted after the branch is deleted? I have not used this workflow yet so I am unfamiliar. Just want to make sure I get this. @ssbarbee

ssbarbee: Hey @Sol3iman so in https://github.com/Voltz-Protocol/voltz-ui/pull/412 -> I added a script that deletes S3 buckets. It seems that we can have maximum of 100 buckets per account active at same time. AWS Amplify creates this buckets for us when a new branch is pushed. Problem here is that after branch is deleted, the hosting env. is also deleted but they don't delete AWS S3 buckets... So this is us fixing an obvious bug from AWS Amplify team. I don't want our build time to increase due to us having ```@aws-sdk/client-s3``` as part of our codebase. It is something you can install globally before you use the cleanup script.

Sol3iman: Makes sense, thank you

Sol3iman (approved):

PR #449, opened 2022-12-21
Added config file for pools

0xZenus:
# Title

Configuration file for Voltz pools

## Description

This PR introduces network-specific configurations for the Voltz pools that include:
   - general show flag (that filters all the pools in general)
   - trader show flag (that filters the trader pools)
   - position construction doesn't build new AMMs but use the same ones as in the pool page
   - forced rollovers - option to force some pools to rollover into specified pools (i.e. bypassing the search algorithm that get the latest pool with the same underlying and rate oracle ID)
   - trader withdrawable flag -- we want to delay settlement and rollovers for traders for the first 48hrs after maturity - so we use this flag to control it.

https://feat-config-file.d2f2fhja7tsifl.amplifyapp.com/#/trader-pools

## Notion link

Please include a Notion link

## Screenshots/videos section

If applicable...


src/hooks/voltz-config/config.ts
```diff
@@ -0,0 +1,271 @@
...
+
+export const networkConfigurations: { [key: string]: NetworkConfiguration } = {
+  mainnet: {
+    factoryAddress: '0x6a7a5c3824508D03F0d2d24E0482Bea39E08CcAF',
+    apply: true,
+    pools: [
+      // aUSDC pools
+      {
+        name: 'aUSDC_v1',
+        id: '0xae16bb8fe13001b61ddb44e2ceae472d6af08755', // vamm address
```
ioanaroceanu2000, line 13: do we have to update the UI every time we create a pool?

src/hooks/voltz-config/config.ts
```diff
@@ -0,0 +1,271 @@
+import { providers } from 'ethers';
+
+import { NetworkConfiguration } from './types';
+
+export const networkConfigurations: { [key: string]: NetworkConfiguration } = {
```
ioanaroceanu2000, line 5: this should be more like `poolConfiguration`

src/hooks/voltz-config/config.ts
```diff
@@ -0,0 +1,271 @@
...
+        },
+        traderWithdrawable: false,
+      },
+    ],
+  },
+
+  goerli: {
+    factoryAddress: '0x9f30Ec6903F1728ca250f48f664e48c3f15038eD',
+    apply: false,
+    pools: [],
```
ioanaroceanu2000, line 243: why empty?

src/hooks/voltz-config/config.ts
0xZenus: yes, we do for mainnet. But we'd have needed anyway because of whitelists.

src/hooks/voltz-config/config.ts
0xZenus: yes, I think networkPoolConfigurations because it's still per network.

ioanaroceanu2000 (approved):

PR #450, opened 2022-12-22
Fix allowance handling

CyclopsRex:
## Description

fix token approvals issues, making use of API 2.0.0
 - dispaly helpful error if approval is goign to fail, warning about tokens like USDT that don't allow approvals on top on non-zero allowances
 - don't require approval if the margin being supplied is less than the existing allowance

Not possible till SDK PR merged and SDK major version is incremented: https://github.com/Voltz-Protocol/v1-sdk/pull/138

## Notion link

https://www.notion.so/voltz/USDT-LP-Approval-bug-73fc8dedf83c4b00baa3c55965edaf28

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
     "@reduxjs/toolkit": "1.9.0",
     "@sentry/react": "7.25.0",
     "@sentry/tracing": "7.25.0",
-    "@voltz-protocol/v1-sdk": "1.23.1",
+    "@voltz-protocol/v1-sdk": "1.24.3",
     "@walletconnect/ethereum-provider": "1.8.0",
     "@walletconnect/web3-provider": "1.8.0",
     "aws-amplify": "4.3.44",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -144,7 +144,7 @@
     "@reduxjs/toolkit": "1.9.0",
     "@sentry/react": "7.25.0",
     "@sentry/tracing": "7.25.0",
-    "@voltz-protocol/v1-sdk": "1.17.0",
+    "@voltz-protocol/v1-sdk": "2.0.0",
```
ssbarbee, line 147: Note: In order to get this version in SDK you need to have at-least one commit containing BREAKING CHANGE: <commit_message> in the SDK repo.

Otherwise you will get minor bump at best.

CyclopsRex: @ssbarbee thanks. I do have that already but shout if you think I've done it wrong. Happy with the change otherwise, in as much as you can review before the SDK is merged?

ssbarbee: I don't see the code for sentry tracking once the issue happens. i think we talked about adding that part as well. Other than than code looks good

CyclopsRex: @ssbarbee The breaking change is this one: https://github.com/Voltz-Protocol/v1-sdk/pull/138/commits/ebff86e6d00ee27f188a55b691e670102887ef39. Is that not the right place for the codeword?

ANdthe sentry code is in the SDK here https://github.com/Voltz-Protocol/v1-sdk/pull/138/commits/0c11d546d08216d9b8b046aa5b38565c7c4e9760#diff-f54f8adcdbe5808dadd709521a89388c17981052003521a2b5342a54850ec3caR1843-R1844 (have tested and it seems to work).

Please resolve conversation once happy and I'll merge the SDK PR.

ssbarbee (approved):

PR #454, opened 2022-12-23
Feat/cv2 s2 start

ioanaroceanu2000:
# Title

Bump SDK with automation of season end switch of subgraph urls.

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Please include an Amplify hosted environment link

## Notion link

Please include a Notion link

## Screenshots/videos section

If applicable...

### DIFF
--- a/src/routes/Profile/helpers.ts
+++ b/src/routes/Profile/helpers.ts
@@ -314,7 +314,8 @@ export const getCommunitySbt = (signer: Signer | null) => {
   const params: SBTConstructorArgs = {
     id: process.env.REACT_APP_COMMUNITY_SBT_ADDRESS || '',
     signer: signer,
-    badgesSubgraphUrl: process.env.REACT_APP_SUBGRAPH_BADGES_URL,
+    currentBadgesSubgraphUrl: process.env.REACT_APP_CURRENT_SUBGRAPH_BADGES_URL,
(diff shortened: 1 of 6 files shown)


src/routes/Profile/helpers.ts
```diff
@@ -314,7 +314,8 @@ export const getCommunitySbt = (signer: Signer | null) => {
   const params: SBTConstructorArgs = {
     id: process.env.REACT_APP_COMMUNITY_SBT_ADDRESS || '',
     signer: signer,
-    badgesSubgraphUrl: process.env.REACT_APP_SUBGRAPH_BADGES_URL,
+    currentBadgesSubgraphUrl: process.env.REACT_APP_CURRENT_SUBGRAPH_BADGES_URL,
```
ssbarbee, line 317: I would rather if we have a better configuration for this. For example: 

`badgesSubgraphUrls: Record<Season['id'], string>` so you define them as following:

```
0: process.env.REACT_APP_SOG_SUBGRAPH_BADGES_URL,
1: process.env.REACT_APP_SO1_SUBGRAPH_BADGES_URL,
2: process.env.REACT_APP_SO2_SUBGRAPH_BADGES_URL,
```

WDYT?

ioanaroceanu2000: we are only interested in current and next. the switch will happen automatically so I don't see the point in storing all of the subgraph urls

ssbarbee (approved):

PR #471, opened 2022-12-29
fix: rm new mellow badges

ioanaroceanu2000:
# Title

Remove spoiler for new mellow badges in the badges list


src/routes/Profile/data/getSeasonBadges/mappers.ts
```diff
@@ -20,10 +20,6 @@ export const BADGE_TYPE_BADGE_VARIANT_MAP: Record<string, BadgeVariant> = {
   '30': 'maxBidding',
   '31': 'topTrader',
   '32': 'mellowLpVault',
-  '39': 'mellowLpVaultTier1',
```
ssbarbee, line 23: Why are we removing them?

PR #502, opened 2023-01-12
feat: verify EIP1271 signatures

CyclopsRex:
# feat: verify EIP1271 signatures

## Description

Smart contracts do not have (known) private keys and cannot sign messages in the same way as other accounts can, but this change allows smart contracts to demonstrate that they accept our terms and conditions using the EIP1271 standard for message "signatures".

This will allow smart contract wallets, such as Gnosis Safe and Argent, to connect to the Voltz webapp after they have accepted the terms and conditions:
 - For Argent, the experience is comparable to Metamask except that, for now, the user will have to sign a message on every login. We should fix this limitation in a separate PR.
 - For Gnosis Safe, acceptance of the terms of service will require an on-chain signature to ensure that enough multisig participants agree, and this will cost a small amount of gas. (There are other ways of signing data in Gnosis Safe that do not involve paying transaction fees, but this is the default approach that users will be taken through if using the Gnosis Safe webapp.)
 - 
## Remaining work

TODO: save signatures so that we don't need to ask for them every time. Signatures should be saved as unchecked and should only grant access in conjunction with an EIP1271 check.


src/contexts/WalletContext/services.ts
```diff
@@ -52,6 +52,53 @@ export const checkForRiskyWallet = async (walletAddress: string) => {
   }
 };
 
+// Check if the provided signature is valid for the smart contract at signer.getAddress(),
```
ssbarbee, line 55: Thanks for the comments! Can we convert them to JS doc type of comments? Documenting the params of the function. You can find examples in the codebase on this. Look at src/utilities folder it has few nicely commented utils.

src/contexts/WalletContext/services.ts
```diff
@@ -52,6 +52,53 @@ export const checkForRiskyWallet = async (walletAddress: string) => {
...
+// using the checks defined in EIP-1271.
+//
+// For simplicity we use the user's Metamask provider and not Alchemy/Infura, though we could change
+// that in future.
+export const isValidEIP1271Signature = async (
```
ssbarbee, line 60: Lets write unit tests for this and keep on improving test coverage of repo 🙏

src/contexts/WalletContext/services.ts
```diff
@@ -52,6 +52,53 @@ export const checkForRiskyWallet = async (walletAddress: string) => {
...
+  } catch (error) {
+    // signature is not valid
+    return false;
+  }
+};
+
+// Check if the message has been "signed" on chain by the smart contract at signer.getAddress(),
+// using the interface defined in EIP-1271. (For at least Gnosis, on-chain signature checks will
+// kick in if signature = 0x.)
+export const isAlreadyEIP1271SignedOnChain = async (
```
ssbarbee, line 93: Lets write unit tests for this and keep on improving test coverage of repo 🙏

CyclopsRex: See https://github.com/Voltz-Protocol/voltz-ui/pull/502#discussion_r1071093981

src/contexts/WalletContext/services.ts
ssbarbee: It is nothing fancy. You can mock the third party services and write scenarios on behaviour based on mock. have a look into the getAMMs.test.ts, there we mock `client.query` which is a third party library (apollo)

ssbarbee: For your case you can mock the parameters you pass to this function and based on the mocked values test the flow.

src/contexts/WalletContext/services.ts
CyclopsRex: Combined the two functions and added JS doc type comments.

src/contexts/WalletContext/services.ts
CyclopsRex: Tried to get jest mocking working with my function this morning and didn't manage it. The mocking you pointed me to (in getAMMs.test.ts) is a simpler case, and testing interactions with metamask and contracts is more complicated - not least because in the Contract class the method we want to mock () only gets added at construction time.

Mocking this stuff should be doable with some more research (e.g. see ideas here: https://github.com/ethers-io/ethers.js/discussions/3095), and I agree we should add more tests, but we need to work out the right pattern and I don't think this PR, which we want to merge to unblock a user who has funds trapped in the platform, is the right place to do that research.

So I'd like to merge this without tests, and we can open a ticket (if we don't already have one) for writing tests of functions that interface with the blockchain. wdyt?

ssbarbee: Sure lets pair on writing tests or I can prepare something and share with you next week 🙏 If the effort now is big we can do it later as you suggest! 

Feel free to resolve this message once you have seen it 🥰

ssbarbee (approved):

PR #570, opened 2023-02-22
Feat/integrate cashflow calculator

avniculae:


src/app/features/swap-form/reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+    setPredictedApyAction: (
+      state,
+      {
+        payload: { value },
+      }: PayloadAction<{
+        value: string;
+      }>,
+    ) => {
+      state.cashflowCalculator.predictedApy = value;
+      if (parseFloat(toUSFormat(value) as string) < 0) {
```
ssbarbee, line 71: Not sure if toUSFormat holds truth here... What are we trying to deduce here? If value inputed in the string is less than 0? Isn't a check if first char of string is '-' enough?

src/app/features/swap-form/reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      const { additionalCashflow, totalCashflow } = amm.getExpectedCashflowInfo({
+        position: state.position as Position | undefined, //TODO Alex
+        fixedTokenDeltaBalance: Number(state.prospectiveSwap.notionalAmount), //TODO Alex
+        variableTokenDeltaBalance: -Number(state.prospectiveSwap.notionalAmount), //TODO Alex
+        variableFactorStartNow: state.cashflowCalculator.variableFactorStartNow,
+        predictedVariableApy: Number(state.cashflowCalculator.predictedApy), //TODO Alex
+      });
+
+      state.cashflowCalculator.additionalCashflow = formatCurrency(
```
ssbarbee, line 97: this formatCurrency should happen in UI layer. In state try to keep as simple value as possible. Lets say number. Then before displaying it (place of usage, format/map it)

src/app/features/swap-form/reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+        additionalCashflow,
+        true,
+        true,
+        2,
+        4,
+      );
+
+      state.cashflowCalculator.totalCashflow = formatCurrency(totalCashflow, true, true, 2, 4);
```
ssbarbee, line 105: this formatCurrency should happen in UI layer. In state try to keep as simple value as possible. Lets say number. Then before displaying it (place of usage, format/map it)

src/app/features/swap-form/reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+      }: PayloadAction<{
+        amm: AMM; //TODO Alex: remove amm and use amm from state
+      }>,
+    ) => {
+      if (state.cashflowCalculator.error) {
+        return;
+      }
+
+      const { additionalCashflow, totalCashflow } = amm.getExpectedCashflowInfo({
+        position: state.position as Position | undefined, //TODO Alex
```
avniculae, line 90: @ssbarbee it complains if I don't cast it here. Do you know why?

src/app/features/swap-form/reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+        4,
+      );
+
+      state.cashflowCalculator.totalCashflow = formatCurrency(totalCashflow, true, true, 2, 4);
+    },
+  },
+  extraReducers: (builder) => {
+    builder
+      .addCase(initialiseCashflowCalculator.pending, (state, {}) => {
+        state.cashflowCalculator.initialised = false;
```
ssbarbee, line 111: to be honest not sure why you don't use the 'idle' | 'success' | 'pending' | 'error' instead of boolean, for the initialised. Right now it is not possible to answer the question is it loading or initial state.

src/app/features/swap-form/reducer.ts
```diff
@@ -0,0 +1,123 @@
...
+      }>,
+    ) => {
+      state.cashflowCalculator.predictedApy = value;
+      if (parseFloat(toUSFormat(value) as string) < 0) {
+        state.cashflowCalculator.error = true;
+      } else {
+        state.cashflowCalculator.error = false;
+      }
+    },
+    refreshCashflows: (
```
ssbarbee, line 77: best practice is to add 'Action' suffix to signal that this is action. Perhaps rename it to updateCashflowCalculatorAction? WDYT

src/utilities/number/index.ts
```diff
@@ -81,7 +81,7 @@ export const removeFormat = (formattedValue: string | undefined): string | undef
  */
 export const toUSFormat = (value: string | undefined): string | undefined => {
   if (withCommaDecimalSeparator()) {
-    return value?.split('.').join('-').split(',').join('.').split('-').join(',');
+    return value?.split('.').join('#').split(',').join('.').split('#').join(',');
```
ssbarbee, line 84: are test passing after this?

src/app/features/swap-form/reducer.ts
ssbarbee: because we should not be keeping classes in state :D Unfortunately current SDK doesn't expose stateless fns around AMMS so we have to do this nasty stuff

src/app/features/swap-form/reducer.ts
avniculae: I think it's cleaner to convert it to a number -- I've fixed the `stringToBigFloat` function and use that instead. See [6b0d2f3](https://github.com/Voltz-Protocol/voltz-ui/pull/570/commits/6b0d2f3843234382bf7a2334f42b64cb5096adce)

src/app/features/swap-form/reducer.ts
avniculae: Addressed [6b0d2f3](https://github.com/Voltz-Protocol/voltz-ui/pull/570/commits/6b0d2f3843234382bf7a2334f42b64cb5096adce)

src/app/features/swap-form/reducer.ts
avniculae: Addressed [6b0d2f3](https://github.com/Voltz-Protocol/voltz-ui/pull/570/commits/6b0d2f3843234382bf7a2334f42b64cb5096adce)

src/app/features/swap-form/reducer.ts
avniculae: Aha, this makes sense, thanks!

src/app/features/swap-form/reducer.ts
avniculae: Addressed [6b0d2f3](https://github.com/Voltz-Protocol/voltz-ui/pull/570/commits/6b0d2f3843234382bf7a2334f42b64cb5096adce)

src/app/features/swap-form/reducer.ts
avniculae: Addressed [6b0d2f3](https://github.com/Voltz-Protocol/voltz-ui/pull/570/commits/6b0d2f3843234382bf7a2334f42b64cb5096adce)

src/utilities/number/index.ts
avniculae: Yes, tests are passing -- using `-` as the auxiliary character would break logic when number are negative

avniculae: Added more tests [f834398](https://github.com/Voltz-Protocol/voltz-ui/pull/570/commits/f834398de9a6df0179b6a7fe7ce15c518c7adaae) to catch this scenarios

ssbarbee (approved):

PR #572, opened 2023-02-24
feat: feed more data into components

avniculae:


src/app/features/swap-form/reducer.ts
```diff
@@ -1,10 +1,14 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
-import { AMM, Position } from '@voltz-protocol/v1-sdk';
+import { AMM, InfoPostSwapV1, Position } from '@voltz-protocol/v1-sdk';
+import { Signer } from 'ethers';
+// eslint-disable-next-line no-restricted-imports
+import { WritableDraft } from 'immer/dist/internal';
```
ssbarbee, line 5: we should not be doing this. There is reason why you should never import from node modules directly. Can you check if it is possible to get this import from another place? 

Use: https://github.com/immerjs/immer/issues/726 it seems a type called `Draft` is exported from `@redux/toolkit`

src/routes/TraderForm/Main/PositionDetails/index.tsx
```diff
@@ -16,8 +23,32 @@ import {
...
+  const fixedRate =
+    infoPostSwap.status === 'success' ? formatNumber(infoPostSwap.value.averageFixedRate) : '--';
+  const variableRate =
+    variableRateInfo.status === 'success' ? formatNumber(variableRateInfo.value) : '--';
+
+  const receivingRate = mode === 'fixed' ? fixedRate : variableRate;
+  const payingRate = mode === 'fixed' ? variableRate : fixedRate;
+
+  const compactNotional = compactFormat(stringToBigFloat(notional.value));
+  let notionalBoxToken = (compactNotional.at(-1) || '').match(/[a-z]/i)
```
ssbarbee, line 41: can you explain this logic a bit? compactNotional.at(-1) what does it mean? It is kinda hard to read this. Also using regex during render is not really optimal. Regex is expensive computation. Perhaps we can write a simpler code. What are you trying to do here actually?

avniculae: `compactNotional.at(-1)` is getting the last character to check if it's a letter or not. Let me write an util function that replaces the regex-based isLetter check with an easier implementation.

ssbarbee: to my mind it kinda feels complex what we are doing to get this 'K', 'M' value and so on

from my analyis problem is 

```
export const compactFormat = (value: number): string => {
  const formatter = Intl.NumberFormat(navigator.language, {
    notation: 'compact',
    maximumFractionDigits: 2,
    minimumFractionDigits: 0,
  });
  return formatter.format(value);
};
```

it doesn't really give us information regarding this letter K
what if we created another function
call it perhaps `compactFormatNumber`

and instead of returning the plain string (formatted) we include the letter as well

```
export const compactFormatNumber = (value: number): string => {
  const formatted = compactFormat(value);
  const lastChar = formatted.at(-1);
  if(isNaN(parseInt(lastChar, 10))) {
    return {
      formatted,
      suffix: lastChar
    }
  }

  return {
    formatted,
    suffix: ''
  }
};
```

now if we try and use this one, we get more information that we can plug into our token components
without having to write complex logic for `isASCIILetter`

we just say const { formatted, suffix } = compatFormatNumber(1111);
// suffix = k, formatted = 1.1
then we pass this to our components

avniculae: Addressed in [3a60d00](https://github.com/Voltz-Protocol/voltz-ui/pull/572/commits/3a60d00a29e7a8c2372ab8cb8aac3d48d7c4c948). However, I would keep the isASCIILetter function in case we need it in the future, since it's already tested. wdyt?

src/app/features/swap-form/reducer.ts
avniculae: [c0724b9](https://github.com/Voltz-Protocol/voltz-ui/pull/572/commits/c0724b9b2c4478849f20f00b94dc25a7f70eb2ae)

src/routes/TraderForm/Main/PositionDetails/index.tsx
ssbarbee: no need to bloat code with stuff we won't need. We will always have reference to this in github history, lets remove it for now.

src/utilities/number/index.ts
```diff
@@ -114,6 +114,34 @@ export const compactFormat = (value: number): string => {
...
+  const formatter = Intl.NumberFormat(navigator.language, {
+    notation: 'compact',
+    maximumFractionDigits: 2,
+    minimumFractionDigits: 0,
+  });
+  const parts = formatter.formatToParts(value);
+  return {
+    compactNumber: parts
+      .filter((part) => part.type !== 'compact')
+      .reduce((compactNumber, part) => compactNumber.concat(part.value), ''),
```
ssbarbee, line 138:
```suggestion
      .join(''),
```

src/utilities/number/index.ts
```diff
@@ -114,6 +114,34 @@ export const compactFormat = (value: number): string => {
...
+    compactSuffix: parts
+      .filter((part) => part.type === 'compact')
+      .reduce((compactSuffix, part) => compactSuffix.concat(part.value), ''),
```
ssbarbee, line 141:
```suggestion
      .join(''),
```

src/utilities/number/index.ts
```diff
@@ -114,6 +114,34 @@ export const compactFormat = (value: number): string => {
   return formatter.format(value);
 };
 
+/**
+ * It takes a number and returns a string using Intl.NumberFormat formatter
+ * it creates compact number format
+ * @param {number} value - number - The number to format.
+ * @returns A string
```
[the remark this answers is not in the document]
avniculae: [bca72f6](https://github.com/Voltz-Protocol/voltz-ui/pull/572/commits/bca72f6ca8b0f4e06eab49265eeff64dbd3feb0a)

ssbarbee (approved):

PR #582, opened 2023-03-05
feat: add notional field for edit position

avniculae:


src/routes/TraderForm/Form/MarginAmountField/index.tsx
```diff
@@ -1,39 +1,46 @@
...
+  useEffect(() => {
+    setLocalMargin(marginAmount.value.toString());
+  }, [marginAmount.value]);
+
   const handleOnChange = useCallback(
     (value?: string) => {
-      if (!value) {
-        return;
-      }
+      setLocalMargin(value ?? null);
```
ssbarbee, line 38: the point of having a localMargin is because we want to debounce the event for dispatching that updates our global store. In this case you are missing the part for debouncing the dispatch. Can you add that as well? And move the dispatch to this debounced function

avniculae: [161562c](https://github.com/Voltz-Protocol/voltz-ui/pull/582/commits/161562c53ef4a56fb707817c0e17c85adcfc6ae0)

ssbarbee (approved):

PR #604, opened 2023-03-21
feat: add GA for swap form

ssbarbee:


src/app/features/swap-form/thunks.ts
```diff
@@ -284,21 +303,38 @@ export const confirmSwapThunk = createAsyncThunk<
...
       margin: getProspectiveSwapMargin(swapFormState),
       fixedLow: 1,
       fixedHigh: 999,
     });
+    pushSwapTransactionSuccessEvent(eventParams);
+    return result;
   } catch (err) {
+    pushSwapTransactionFailedEvent({
+      ...eventParams,
+      errorMessage: extractError(err),
```
ioanaroceanu2000, line 336: is this the message we show to users? asking because in GA we need a short error message not something like sentry

src/utilities/googleAnalytics/index.ts
```diff
@@ -21,14 +19,17 @@ export type DataLayerEventPayload = {
...
-  agent?: Agents;
+  agent?: 'Fixed Trader' | 'Variable Trader' | 'Liquidity Provider';
 };
 
 type DataLayerEventName =
   | 'expectedApy_change'
   | 'notional_change'
   | 'margin_change'
+  // TODO: Soon deprecated
   | 'leverage_change'
```
ioanaroceanu2000, line 30: you can remove this now tbh

src/app/features/swap-form/thunks.ts
ssbarbee: It is short message

src/utilities/googleAnalytics/index.ts
ssbarbee: sure!

ioanaroceanu2000 (approved):

PR #615, opened 2023-03-28
feat: add more trader form uts

avniculae:


src/app/features/swap-form/utils/getExistingPositionMode/index.ts
```diff
@@ -0,0 +1,11 @@
+import { Draft } from '@reduxjs/toolkit';
+
+import { SliceState } from '../../reducer';
+
+export const getExistingPositionMode = (state: Draft<SliceState>) => {
+  if (state.position.status !== 'success' || !state.position.value) {
+    return null;
+  }
+
+  return state.position.value.positionType === 1 ? 'fixed' : 'variable';
```
ssbarbee, line 10: is it not possible to get some sort of enum from SDK for this check here? it feels really random number 1

ssbarbee (approved):

